PR #3, opened 2017-04-06
Handling case where destination host (ftp server) is not local host

loerum:
Few changes for:
* Handling case where destination host (ftp server) is not local host.
* Introducing BIG_TIMEOUT.
* A few more LOGGER.debug messages for novices.


mraspaud:
Looks good ! :+1:

PR #7, opened 2017-09-29
Introducing heartbeat monitor

loerum:
Introducing:

* Restart of clients listeners (by a threading.Event).
* A heartbeat monitor, controlled by a `heartbeat_alarm_scale` option (in need of a better name !).
  * Heartbeat monitor will do nothing if scale option not set or is zero.
  *  If set, monitor will generate an alarm (restart) event if no heartbeat received after
    `heartbeat_alarm_scale` * `interval between heartbeats` seconds.

### DIFF
--- a/pytroll_fu/client.py
+++ b/pytroll_fu/client.py
@@ -28,7 +28,7 @@ import sys
 import time
 from collections import deque
 from ConfigParser import ConfigParser
-from threading import Lock, Thread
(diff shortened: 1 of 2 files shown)


loerum:
Hmm ... "New errors were found" ... fx. typo in `move_it_mirror.local_reload_config(finename)` ... look like it's coming from develop branch ?

PR #8, opened 2017-11-17
Introducing 'info' request

loerum:
Implementing a "info" request to a move it server. For a request with topic "topic"
and type "info" the server will reply with a list of published files belonging to 
topics starting with "topic".

Notes (and room for improvements):
* Server will keep cache of published files (like client).
* Currently there is some (arbitrary) limit on cache size and amount of data to be send.
* The cache is not persistent. This is kind of OK, since a server on start up, will publish
  files in "origin" (not quite correct, the original order of published files are lost).
* Introducing bin/move_it_req.py to send request (ping or info) to a server.
* No time stamp on published files, so an info request can only specify a max count of files
  (not what has been published since yesterday at 12).

### DIFF
--- a/bin/move_it_req.py
+++ b/bin/move_it_req.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2017
+
+# Author(s):
+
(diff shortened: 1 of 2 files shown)


mraspaud:
@loerum lgtm, can I merge this ?

PR #11, opened 2018-02-22
Added argument option disable-backlog

TAlonglong:
This turn off publishing of files on disk at startup/restart. This can be usefull to avoid a lot of old files published at startup

I have only put this disable-backlog at startup. At reload from signal or inotify event don't apply this flag


TAlonglong:
One effect of this will be all files scheduled for removal will not be removed.

PR #13, opened 2018-08-03
Improve error handling in SCPMover

TAlonglong:
If something wrong happens with the scp.put the ssh and scp object is left open. They need to be closed.

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -657,11 +657,14 @@ class ScpMover(Mover):
         LOGGER.debug('dest path %s ', os.path.dirname(self.destination.path))
         LOGGER.debug('origin %s ', self.origin)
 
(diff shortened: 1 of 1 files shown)


trollmoves/server.py
```diff
@@ -657,11 +657,14 @@ def copy(self):
...
-
-        scp.close()
-        ssh.close()
+        try:
+            scp = SCPClient(ssh.get_transport())
+            scp.put(self.origin, self.destination.path)
+        except Exception as e:
+            LOGGER.error("Something went wrong with scp: " + str(e))
+        finally:
+            scp.close()
```
mraspaud: if the `SCPClient` call fails, the `scp` object isn't creates, and the subsequent call to `close` will crash...

PR #15, opened 2018-08-10
Make ftp connections persistent for 30 seconds

mraspaud:


mraspaud:
yes, exactly. The error is then sent as a message so that the client is aware of the problem, and iirc, the client will go on to the next server providing the file to retrieve it.

mraspaud:
@TAlonglong Did you have a chance to test this ?

PR #29, opened 2019-05-31
Add polling scan from server, xrit unpack on client and other minor features

vgiuffrida:
Some new features added:

Client:
- xrit decompression can be executed by the client.
xrit functions have been moved from server to utils.py script since they are used also from the client.
- reference file mode: if configured, generate a reference file (REF file) when a complete set of data is processed. The REF file is a text file containing information on where the complete set of data has been decompressed.
- procfile_log: generation of a log file containing the leatest segments processed by the client. Used to avoid reprocessing of old data.

Server:
- Add new polling scanmode: if configure, polling mode is used to detect new files instead of pynotify

Documentation:
- Update docstring with the new configurable parameters
- Add two example config files containing the new parameters

### DIFF
--- a/bin/move_it_client.py
+++ b/bin/move_it_client.py
@@ -106,7 +106,7 @@ Others configuration values:
                  reference file content:
                                           [REF]
                                           SourcePath = referenced_directory_contains_unpacked_data
-                                          FileName = name_of_ref_file
+                                          FileName = file that triggered generation of ref file
 
 
 Logging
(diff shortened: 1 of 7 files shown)


mraspaud:
@vgiuffrida Thank you for the comprehensive PR.

It feels like all the features you add would need their own PR. Could you split this onto differents ones ? That would make it easier to discuss and make sure it doesn't break existing functionality.

vgiuffrida:
Thank you for reviewing this.

Ok, I'll close this PR and we'll split changes on multiple PRs

PR #34, opened 2019-06-25
Make it possible to provide per-provider topic

pnuu:
With this PR it is possible to provide a topic specific to each provider. If `topic` (for client) or `client_topic` (for mirror) is defined, it will be subscribed on every provider.

### DIFF
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -106,15 +106,19 @@ class Listeners(object):
     def __init__(self, callback, client_topic, providers, **attrs):
         self.listeners = []
         if client_topic is None:
-            topics = []
+            client_topic = []
         else:
-            topics = [client_topic]
+            client_topic = [client_topic]
+
         for provider in providers.split():
+            topic = client_topic
             if '/' in provider:
                 parts = provider.split('/', 1)
                 provider = parts[0]
-                topics.append('/' + parts[1])
-            self.listeners.append(Listener('tcp://' + provider, topics,
+                topic = ['/' + parts[1]]
+                LOGGER.info("Using provider-specific topic %s for %s"
+                            topic, provider)
+            self.listeners.append(Listener('tcp://' + provider, topic,
                                            callback, **attrs))
 
     def start(self):
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -90,6 +90,8 @@ class MoveItMirror(MoveItBase):
             LOGGER.debug('Sending %s', str(new_msg))
             send(str(new_msg))
 
+        if "client_topic" not in attrs:
+            attrs["client_topic"] = None
         listeners = Listeners(publish_callback, **attrs)
 
         return listeners, foo
@@ -103,8 +105,16 @@ class Listeners(object):
 
     def __init__(self, callback, client_topic, providers, **attrs):
         self.listeners = []
+        if client_topic is None:
+            topics = []
+        else:
+            topics = [client_topic]
         for provider in providers.split():
-            self.listeners.append(Listener('tcp://' + provider, [client_topic],
+            if '/' in provider:
+                parts = provider.split('/', 1)
+                provider = parts[0]
+                topics.append('/' + parts[1])
+            self.listeners.append(Listener('tcp://' + provider, topics,
                                            callback, **attrs))
 
     def start(self):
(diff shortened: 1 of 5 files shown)


trollmoves/client.py
```diff
@@ -407,6 +407,10 @@ def reload_config(filename, chains, callback=request_push, pub_instance=None):
             if val.get("heartbeat", False):
                 topics.append(SERVER_HEARTBEAT_TOPIC)
             for provider in chains[key]["providers"]:
+                if '/' in provider.split(':')[-1]:
```
mraspaud: shouldn't this be:
```suggestion
                if '/' in provider.split(':')[1]:
```

trollmoves/client.py
```diff
@@ -407,6 +407,10 @@ def reload_config(filename, chains, callback=request_push, pub_instance=None):
...
+                    parts = urlparse(provider)
+                    provider = parts.scheme + '://' + parts.netloc
```
mraspaud: use `urlunparse` for this.

examples/move_it_mirror.ini
```diff
@@ -62,3 +62,24 @@ topic = /1b/hrit-segment/0deg
...
 # delete = True
+
+[eumetcast-hrit-0deg_scp_per_server_topic]
+# client part, the incoming data, providers with topics
+providers = satmottag2:9011/topic/1 satmottag:9011/topic/2
+login = user
+# Destination where the data are saved locally
+destination = scp:///tmp/bar/
+# Topics listened for advertised incoming data from all providers. Optional.
+client_topic = /1b/hrit-segment/0deg
```
mraspaud: I'm not clear on the difference between this and what is specified in `providers`

bin/move_it_mirror.py
```diff
@@ -103,8 +105,16 @@ class Listeners(object):
...
+        if client_topic is None:
+            topics = []
+        else:
+            topics = [client_topic]
         for provider in providers.split():
-            self.listeners.append(Listener('tcp://' + provider, [client_topic],
+            if '/' in provider:
+                parts = provider.split('/', 1)
+                provider = parts[0]
+                topics.append('/' + parts[1])
```
mraspaud: So one provider can have multiple topics ?

mraspaud:
Many questions here. I guess we need to write some documentation about this. Could you write a paragraph or two about operating a mirror with multiple topics ?

trollmoves/client.py
pnuu: Here the `provider` has `tcp://` prefixed, so the last part of the provider string needs to be checked for the topic.

trollmoves/client.py
pnuu: Using `urlunparse` would be unnecessarily complicated as the attributes can't be modified, and the topic (`parts.path`) needs to be removed from the provider.

examples/move_it_mirror.ini
pnuu: `client_topic` can be considered as a global default topic to subscribe to. So some of the providers may have a specific topic, and for the others the global one would still be available.

bin/move_it_mirror.py
pnuu: The underlying `NSSubscriber` used by the client uses list of topics, so thought that would be the way to go. Do you think the `provider` topic, if specified, should overwrite the global topic?

pnuu:
The mirror is still a bit of a mystery to me, but what I was thinking was:

- User has own servers, call them `own_a` and `own_b` serving the same data and acting as each others' backups, both servers using the same topic
- The `mirror` is used as a gateway to other external servers that can't be connected directly, and they might have different topics to what the local servers use
- The user wants to have the data from `own_a` and `own_b` directly without the data first being transferred to `mirror`

In this case the user can configure their client to listen to the topic used by local servers and the one published by the mirror, and wouldn't get duplicate data and would get the data with minimum delay as long as the local servers receive the data.

trollmoves/client.py
mraspaud: Then you need to make sure that there is a scheme in the provider url
Maybe using urlsparse could help making sure we look at the right part ?

trollmoves/client.py
mraspaud: I insist :) I think it'll make the code more robust.

bin/move_it_mirror.py
mraspaud: yes, otherwise it might lead to unwanted topics being monitored

pnuu:
Ok, I now use more `urlparse`/`urlunparse` to format the provider URI.

PR #36, opened 2019-08-09
Prevent concurrent transfers when multiples sources are available

mraspaud:
The current move_it_client doesn't discriminate when several sources are available for a file and request pushes for each of them. This leads to problems as the same file is written to from multiple sources.

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -314,6 +314,7 @@ def replace_mda(msg, kwargs):
             msg.data[key] = replacement[msg.data[key]]
     return msg
 
+
 def send_request(msg, req, timeout):
     """Send a request for push."""
     LOGGER.debug("Send and recv timeout is %.2f seconds", timeout)
@@ -323,9 +324,8 @@ def send_request(msg, req, timeout):
     return requester.send_and_recv(req, timeout=timeout), hostname
 
 
-def send_ack(msg, publisher, timeout):
+def send_ack(msg, timeout):
     """Send an ACK (no push required)."""
-    resend_if_local(msg, publisher)
     req = Message(msg.subject, 'ack', data=msg.data)
     LOGGER.debug("Sending: %s", str(req))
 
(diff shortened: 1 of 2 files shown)


trollmoves/client.py
```diff
@@ -312,57 +315,116 @@ def replace_mda(msg, kwargs):
...
+    with ongoing_transfers_lock:
+        if huid in ongoing_transfers:
+            ongoing_transfers[huid].append(msg)
+            return
+        else:
+            ongoing_transfers[huid] = [msg]
+
+    if already_received(msg):
+        timeout = float(kwargs["req_timeout"])
+        resend_if_local(msg, publisher)
```
mraspaud: @pnuu @TAlonglong would you be ok if I removed this line ? This basically resends the message if the file is in the file_cache (meaning it has already been processed) and that the files are pointing to the local machine. I don't think this is actually needed. What do you think ?

pnuu: I'm ok with that. But note that I have only ran few tests so far, so don't put too much weight on my thougts :grin: Anyway, sounds like removing it won't be affecting our production in the future either.

TAlonglong: Sorry, I dont think I get the logic. Why would you resend the message if its already received and is local? Its a lots of thinks I dont get in trollmove, so maybe I just stay away :-)

mraspaud: No, you are right. I think it's some legacy stuff that I forgot to remove. I'll clean this up now.

mraspaud:
@TAlonglong we use move_it_client to transfer SDR file collections too.

TAlonglong:
> @TAlonglong we use move_it_client to transfer SDR file collections too for example.

Ah, ok. Nice to know. I only have used it to transfere 1 file at the time.

mraspaud:
@pnuu do you want to review this or should I just merge it ?

pnuu:
Go ahead and merge. Just got back from holidays, and in addition need to move to another office room, so won't have much time in the few next days. I'll complain later if something doesn't work :grinning:

PR #37, opened 2019-09-11
Add dispatcher utility

mraspaud:
Add the dispatcher utility.

```yaml
client1:
  host: ftp://ftp.client1.com
  connection_parameters:
    credential_file: .netrc
  filepattern: '{platform_name}_{start_time}.{format}'
  directory: /input_data/{sensor}
  dispatch:
    - topics:
        - /level2/viirs
        - /level2/avhrr
      conditions:
        # key matches metadata items or provides default
        - product: [green_snow, true_color]
          sensor: viirs
        - product: [green_snow, overview]
          sensor: avhrr
          # special section "except" for negating
          except:
            platform_name: NOAA-15
    - topics:
        - /level3/cloudtype
      directory: /input/cloud_products
      conditions:
        - area: omerc_bb
          daylight: '<30'
          coverage: '>50'
```


pnuu:
I'll give some more comments tomorrow morning.

trollmoves/dispatcher.py
```diff
@@ -0,0 +1,432 @@
...
+      connection_parameters:
+        connection_uptime: 60
+      filepattern: '{platform_name}_{start_time}.{format}'
+      directory: /input_data/{sensor}
+      aliases:
+        product:
+          natural_color: dnc
+          overview: ovw
+      dispatch:
+        - topics:
```
pnuu: Is there a need to match other things, like originating host or `service`? Looking at the very generic example topics, I could see a need to pre-filter the messages with something more, e.g. when there are development streams being tested.

trollmoves/dispatcher.py
```diff
@@ -0,0 +1,432 @@
...
+
+It listens to posttroll messages to find out what to dispatch.
+
+
+Format of the configuration file
+================================
+
+.. highlight:: yaml
+
+Example config::
```
pnuu: Could we have the example config also in the `examples/` directory?

trollmoves/dispatcher.py
```diff
@@ -0,0 +1,432 @@
...
+
+    client1:
```
pnuu: I believe this is just a generic name for the configuration section? Could it be renamed to something like `target_host_1_config` to distinguish from `client` side of `trollmoves`.

trollmoves/dispatcher.py
```diff
@@ -0,0 +1,432 @@
...
+      host: ftp://ftp.client1.com
```
pnuu: Maybe add one level more here, e.g.:
```yaml
  target_host:
    uri: ftp://ftp.client1.com
    connection_parameters:
      credential_file: .netrc
    filepattern: '{platform_name}_{start_time}.{format}'
    directory: /input_data/{sensor}
```
Just to make it more clear these will are for the target system, not e.g. `filepattern` of the incoming files.

pnuu:
Could there be a `condition` for filename patterns? I'm thinking about messages coming from `trollmoves` transfers, which might need additional transfers to servers without `trollmoves` client capabilities, like external customers. Although now that I look at those messages I think there might be enought information even without this feature :thinking:

mraspaud:
I was hoping to be free from the filenames and just work with metadata...

trollmoves/dispatcher.py
mraspaud: Good point. Something like this ?
```suggestion
    target1:
```

trollmoves/dispatcher.py
mraspaud: Yep, I'll put it there

trollmoves/dispatcher.py
mraspaud: It's really up to the user to define what topics they would like to use. Do you want me to give a more complex example ?
Regarding service or host, for me that would be matched with the metadata somehow, but topics should work too.

trollmoves/dispatcher.py
mraspaud: so you would like:
```yaml
client1:
  target_host:
    uri: ftp://ftp.client1.com
``` 
etc ?
I'm not sure I like it, it adds an 'unused' level, a level with just one item.

trollmoves/dispatcher.py
pnuu: I'll be having a very detailed topics, so I'm fine with this. It was just a thought that someone might want to have only data coming from local reception to be forwarded even if there are other sources for the same products with the same topics.

trollmoves/dispatcher.py
pnuu: Yes, that's more clear.

trollmoves/dispatcher.py
pnuu: Yes, like that. If not, I think there should anyway be a clear separation between the message metadata bits and items relevant to the target host. Oh, if `dispatch` is renamed to `metadata_filters`, and maybe `filepattern` to `target_filepattern` and `directory` to `target_directory` that should do the trick.

pnuu:
Isn't `filename` (or `uid`/`uri`) also metadata? :grin: Anyway, this can also be handled by having specific topics.

trollmoves/dispatcher.py
mraspaud: `dispatch_configs` will be used as per slack conversation

pnuu:
LGTM, feel free to merge if you don't find anything to change.

mraspaud:
Thanks. I'll test my current dispatch chains with the modifications to make sure it still works

PR #41, opened 2019-09-27
Accept log configs in dispatcher

mraspaud:
This uses yaml since dictConfig is now the recommended way to provide 
configuration to logging, and yaml examples are provided in the python 
documentation.

The long term plan is to accept log configs in other scripts too.


mraspaud:
1. right now, only dispatch support passing a log config file, but the plan is to expand this to the other scripts
2. the monitoring isn't published yet (we have a pytroll-monitor package locally), but I plan to publish it soon. Just have to remove the sensitive bits.
What do you want me to do ?

PR #42, opened 2019-10-04
Hot-spare client(s)

pnuu:
With this PR it is possible to have one or more hot-spare clients besides the primary client, so it is possible to update things without disturbing the transfers.

### DIFF
--- a/examples/move_it_client.ini
+++ b/examples/move_it_client.ini
@@ -44,3 +44,19 @@ login = user
 # topic = /1b/hrit-segment/0deg
 # Port for file requests: 0 = random
 publish_port = 0
+
+# Example acting as a hot spare
+[eumetcast_hrit_0deg_scp]
+# Servers to listen, <server>:<port>
+# Use the port number defined with the -p flag for server or mirror
+providers = satmottag2:9010 satmottag:9010 explorer:9010
+# Local destination for the data using SCP
+destination = scp:///tmp/foo
+# Login credentials for local SSH server.  Using keys, so no password given
+login = user
+# Topic to follow
+topic = /1b/hrit-segment/0deg
+# Port for file requests: 0 = random
+publish_port = 0
+# Wait a tiny bit of time when being a hot spare
+processing_delay = 0.01
--- a/examples/move_it_client.ini
+++ b/examples/move_it_client.ini
@@ -60,5 +60,5 @@ login = user
 topic = /1b/hrit-segment/0deg
 # Port for file requests: 0 = random
 publish_port = 0
-# Wait a tiny bit of time when being a hot spare
+# Wait a tiny bit of time when being a hot spare.  Delay is given in seconds.
 processing_delay = 0.02
--- a/examples/move_it_client.ini
+++ b/examples/move_it_client.ini
@@ -49,7 +49,9 @@ publish_port = 0
 [eumetcast_hrit_0deg_scp]
 # Servers to listen, <server>:<port>
 # Use the port number defined with the -p flag for server or mirror
-providers = satmottag2:9010 satmottag:9010 explorer:9010
+# The last one is another client that is the primary client, and the messages
+# coming from there need to be also handled
+providers = satmottag2:9010 satmottag:9010 explorer:9010 primary_client
 # Local destination for the data using SCP
 destination = scp:///tmp/foo
 # Login credentials for local SSH server.  Using keys, so no password given
@@ -59,4 +61,4 @@ topic = /1b/hrit-segment/0deg
 # Port for file requests: 0 = random
 publish_port = 0
 # Wait a tiny bit of time when being a hot spare
-processing_delay = 0.01
+processing_delay = 0.02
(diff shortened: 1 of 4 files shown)


pnuu:
To summarize the operation principle:
1. two clients, `primary` and `spare`
2. both clients listen to same providers
3. `spare` also listens to `primary`
4. `spare` waits a bit before acting on the messages coming from providers
5. `primary` publishes a message saying "I'm working on this"
6. `spare` receives the "I'm working on this" message, adds the message to ongoing transfers
7. waiting started in step 4. timeouts, and `spare` sees that there's an ongoing transfer, so does nothing
8. after `primary` completes the transfer, it sends a message saying "ok, I got the file" and `spare` clears the transfer from its local queue and adds the filename to its cache

examples/move_it_client.ini
```diff
@@ -44,3 +44,21 @@ login = user
...
+# Local destination for the data using SCP
+destination = scp:///tmp/foo
+# Login credentials for local SSH server.  Using keys, so no password given
+login = user
+# Topic to follow
+topic = /1b/hrit-segment/0deg
+# Port for file requests: 0 = random
+publish_port = 0
+# Wait a tiny bit of time when being a hot spare
+processing_delay = 0.02
```
mraspaud: Is this seconds ?

trollmoves/client.py
```diff
@@ -179,21 +174,45 @@ def run(self):
...
+                        # the transfers are not finished on primary
+                        # client and are not cleared
+                        LOGGER.debug("Primary client published 'push'")
+                        add_to_ongoing(msg)
+
+                    # Handle public "ack" messages as a hot spare client
+                    if msg.type == "ack":
+                        LOGGER.debug("Primary client finished transfer")
+                        _ = add_to_file_cache(msg)
+                        _ = clean_ongoing_transfer(get_msg_uid(msg))
```
mraspaud: why the `_`s ?

trollmoves/client.py
```diff
@@ -503,9 +554,10 @@ def reload_config(filename, chains, callback=request_push, pub_instance=None):
     # disable old chains
 
     for key in (set(chains.keys()) - set(new_chains.keys())):
-        for provider, listener in chains[key]["providers"].items():
+        for provider in chains[key]["providers"]:
+            listener = chains[key]["listeners"][provider]
             listener.stop()
-            del chains[key]["providers"][provider]
+            del chains[key]["listeners"][provider]
```
mraspaud: Nice catch !

trollmoves/client.py
```diff
@@ -179,21 +174,45 @@ def run(self):
...
+                    # Handle public "ack" messages as a hot spare client
+                    if msg.type == "ack":
+                        LOGGER.debug("Primary client finished transfer")
+                        _ = add_to_file_cache(msg)
+                        _ = clean_ongoing_transfer(get_msg_uid(msg))
+
+                    # If this is a hot spare client, wait for a while
+                    # for a public "push" message which will update
+                    # the ongoing transfers before starting processing here
+                    time.sleep(float(self.ckwargs.get("processing_delay", 0.0)))
```
mraspaud: I see a potential bug in here, if many files are being transfered at once (eg when restarting move it server and trying to catch up on accumulated files), stince the sleeps will accumulate.

As an alternative, I would propose using a `CTimer` (from `trollmoves.movers`), like this:
```python
if self.ckwargs.get("processing_delay`, 0):
    timer = CTimer(self.callback, args=[msg] + self.cargs, kwargs=self.ckwargs)
    timer.start()
else:
    self.callback(msg, *self.cargs, **self.ckwargs)
```
You probably want to save the `timer` so that you can interrupt it in `clean_ongoing_transfer`

mraspaud:
Also, I know we don't have tests for these parts of the server, but is there something of the new feature you could test without having to write tests for everything ?

examples/move_it_client.ini
pnuu: Yes. I'll add the unit to the comment.

trollmoves/client.py
pnuu: Just to show that the functions return a value, but we don't care about them here. Could have also been a flake8 warning about that, but not sure.

trollmoves/client.py
pnuu: Hmm. Right. I always disable the re-send option on the `server`, so didn't think of that. Also, as in the future (when this hot-spare feature is taken to use) we'll have two independent `server` instances hosting the files, so we shouldn't need resending. I'll have a look at the `CTimer`.

pnuu:
The `server` side doesn't need to know anything about the spare `client` instances. I think a realistic testing of the hot spare could be done, but it might be rather complicated:
- start a nameserver
- start a server with `mock`'d file watcher and transfer parts
- start two clients (primary and spare)
- send a message as a server
- check that only one of the clients made a request
- ...

pnuu:
Ok, I adopted the `CTimer`, and running `server`, `primary` client and `spare` client on my localhost things seem to still work :-)

PR #44, opened 2019-10-17
Expose ListenerContainer config items

pnuu:
This PR exposes config ListenerContainer config items  `nameserver`, `addresses` and `services`.

### DIFF
--- a/examples/dispatch.yaml
+++ b/examples/dispatch.yaml
@@ -4,6 +4,12 @@ target1:
     connection_uptime: 60
   filepattern: '{platform_name}_{start_time}.{format}'
   directory: /input_data/{sensor}
+  subscribe_addresses:
+    - tcp://127.0.0.1:40000
+  nameserver: 127.0.0.1
+  subscribe_services:
+    - service_name_1
+    - service_name_2
   aliases:
     product:
       natural_color: dnc
--- a/examples/dispatch.yaml
+++ b/examples/dispatch.yaml
@@ -4,12 +4,15 @@ target1:
     connection_uptime: 60
   filepattern: '{platform_name}_{start_time}.{format}'
   directory: /input_data/{sensor}
-  subscribe_addresses:
-    - tcp://127.0.0.1:40000
-  nameserver: 127.0.0.1
-  subscribe_services:
-    - service_name_1
-    - service_name_2
+  # Optional direct subscriptions
+  # subscribe_addresses:
+  #   - tcp://127.0.0.1:40000
+  # Nameserver to connect to. Optional. Defaults to localhost
+  # nameserver: 127.0.0.1
+  # Subscribe to specific services. Optional. Default: connect to all services
+  # subscribe_services:
+  #   - service_name_1
+  #   - service_name_2
   aliases:
     product:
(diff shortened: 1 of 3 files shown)


pnuu:
Ok, I've tested this in test production, and it works nicely.

examples/dispatch.yaml
```diff
@@ -4,6 +4,12 @@ target1:
     connection_uptime: 60
   filepattern: '{platform_name}_{start_time}.{format}'
   directory: /input_data/{sensor}
+  subscribe_addresses:
+    - tcp://127.0.0.1:40000
+  nameserver: 127.0.0.1
+  subscribe_services:
+    - service_name_1
+    - service_name_2
```
mraspaud: Could these be marked as optional in the example too ?

trollmoves/dispatcher.py
```diff
@@ -39,6 +39,12 @@
         connection_uptime: 60
       filepattern: '{platform_name}_{start_time}.{format}'
       directory: /input_data/{sensor}
+      subscribe_addresses:
+        - tcp://127.0.0.1:40000
+      nameserver: 127.0.0.1
+      subscribe_services:
+        - service_name_1
+        - service_name_2
```
mraspaud: Could these be documented as optional already here ?

trollmoves/dispatcher.py
```diff
@@ -259,7 +268,13 @@ def update_config(self, new_config):
                     # FIXME: make sure to get the last messages though
                     self.listener.stop()
                 self.config = new_config
-                self.listener = ListenerContainer(topics)
+                addresses = client_config.get('subscribe_addresses', None)
+                nameserver = client_config.get('nameserver', 'localhost')
+                services = client_config.get('subscribe_services', '')
```
mraspaud: Don't we want to use the posttroll defaults here ? (instead of redefining them)

examples/dispatch.yaml
pnuu: Done

trollmoves/dispatcher.py
pnuu: Done

trollmoves/dispatcher.py
pnuu: Those are the default values `ListenerContainer` has. Don't know how to get the default values from `ListenerContainer` in the case they are not defined in the config by the user. Or is there a way to say `nameserver = client_config.get('nameserver', default_value_of_nameserver_in_ListenerContainer)` I'm unaware of?

examples/dispatch.yaml
pnuu: Done

trollmoves/dispatcher.py
pnuu: Done

trollmoves/dispatcher.py
pnuu: @mraspaud any additional comments/suggestions to this?

mraspaud: well, we could go introspective, but I don't think it's necessary. This will do for now.

PR #45, opened 2019-10-17
Metno additions

TAlonglong:
A few additions I have added locally

- MEOS creates tmp files which is removed at once. Need to handle this in the server
- MEOS init the hrpt file with 0 size. Need to handle this in the server
- Possible to detect new directories from the servers
  eg in config file like this:
            origin=/data/pytroll/aapp-outdir-ears/{platform_name_dir}_{start_time_dir:%Y%m%d_%H%M}_{orbit_number_dir:05d}/{sensor}_{platform_name}_{start_time:%Y%m%d_%H%M}_{orbit_number:05d}.{data_processing_level:3s}
           origin_inotify_base_dir_skip_levels=-2
- Add configurable remove delay in server
- Add possible trollsift destination in client

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -237,7 +237,7 @@ def create_push_req_message(msg, destination, login):
     fake_req = Message(msg.subject, 'push', data=msg.data.copy())
     try:
         _destination = compose(destination, msg.data)
-    except:
+    except Exception as exp:
         _destination = destination
     duri = urlparse(_destination)
     scheme = duri.scheme or 'file'
@@ -293,7 +293,7 @@ def unpack_and_create_local_message(msg, local_dir, unpack=None, delete=False):
 def make_uris(msg, destination, login=None):
     try:
         _destination = compose(destination, msg.data)
-    except:
+    except Exception as exp:
         _destination = destination
     duri = urlparse(_destination)
     scheme = duri.scheme or 'ssh'
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -237,7 +237,7 @@ def create_push_req_message(msg, destination, login):
     fake_req = Message(msg.subject, 'push', data=msg.data.copy())
     try:
         _destination = compose(destination, msg.data)
-    except Exception as exp:
+    except Exception:
         _destination = destination
     duri = urlparse(_destination)
     scheme = duri.scheme or 'file'
@@ -293,7 +293,7 @@ def unpack_and_create_local_message(msg, local_dir, unpack=None, delete=False):
 def make_uris(msg, destination, login=None):
     try:
         _destination = compose(destination, msg.data)
-    except Exception as exp:
+    except Exception:
         _destination = destination
     duri = urlparse(_destination)
     scheme = duri.scheme or 'ssh'
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -237,8 +237,15 @@ def create_push_req_message(msg, destination, login):
     fake_req = Message(msg.subject, 'push', data=msg.data.copy())
     try:
         _destination = compose(destination, msg.data)
-    except Exception:
-        _destination = destination
+    except KeyError as ke:
+        LOGGER.error("Format identifier is missing from the msg.data: %s", str(ke))
+        raise
+    except ValueError as ve:
+        LOGGER.error("Type of format identifier doesn't match the type in m msg.data: %s", str(ve))
+        raise
+    except AttributeError as ae:
+        LOGGER.error("msg or msg.data is None: %s", str(ae))
+        raise
     duri = urlparse(_destination)
     scheme = duri.scheme or 'file'
     dest_hostname = duri.hostname or socket.gethostname()
@@ -293,8 +300,15 @@ def unpack_and_create_local_message(msg, local_dir, unpack=None, delete=False):
 def make_uris(msg, destination, login=None):
     try:
         _destination = compose(destination, msg.data)
-    except Exception:
-        _destination = destination
+    except KeyError as ke:
+        LOGGER.error("Format identifier is missing from the msg.data: %s", str(ke))
+        raise
+    except ValueError as ve:
+        LOGGER.error("Type of format identifier doesn't match the type in m msg.data: %s", str(ve))
+        raise
+    except AttributeError as ae:
+        LOGGER.error("msg or msg.data is None: %s", str(ae))
+        raise
     duri = urlparse(_destination)
     scheme = duri.scheme or 'ssh'
     dest_hostname = duri.hostname or socket.gethostname()
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -235,18 +235,7 @@ def resend_if_local(msg, publisher):
 
 def create_push_req_message(msg, destination, login):
     fake_req = Message(msg.subject, 'push', data=msg.data.copy())
-    try:
-        _destination = compose(destination, msg.data)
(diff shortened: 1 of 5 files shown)


TAlonglong:
Hm, destination trollsift(or more correctly destination directory does not exist) only works when used with ftp or sftp as far as I can see. Not with scp.

TAlonglong:
OK, so I think I have added what I need. But I'm sure we can find a solution if you find this a problem.

@pnuu @martin can you have a look when you can find some spare time?

trollmoves/client.py
```diff
@@ -234,7 +235,11 @@ def resend_if_local(msg, publisher):
 
 def create_push_req_message(msg, destination, login):
     fake_req = Message(msg.subject, 'push', data=msg.data.copy())
-    duri = urlparse(destination)
+    try:
+        _destination = compose(destination, msg.data)
+    except Exception:
```
pnuu: I think this exception can only be:
 - `KeyError` when format identifier is missing from `msg.data`
 - `ValueError` when type of format identifier doesn't match the type in `msg.data`
 - `AttributeError` when `msg` is None (I'm not sure this can even happen)
Anyway, the exlpicit errors should be used.

trollmoves/movers.py
```diff
@@ -329,6 +329,12 @@ def copy(self):
             ssh_connection.close()
             raise
 
+        try:
+            ssh_connection.exec_command("mkdir -pv {}".format(self.destination.path))
```
pnuu: This is a nice addition! Should there be a explicit config option which, if set to `True` (or equivalent) enables the feature? In some cases this as a default behaviour can hide some quite significant errors :-)

trollmoves/movers.py
```diff
@@ -329,6 +329,12 @@ def copy(self):
...
+        except Exception as err:
```
pnuu: If possible, add some explicit errors that can happen here.

trollmoves/server.py
```diff
@@ -67,15 +67,17 @@ class ConfigError(Exception):
...
         self.queue = Queue()
         self.timer = None
         self.loop = True
+        self._attrs = attrs
 
     def add(self, filename):
         LOGGER.debug('Scheduling %s for removal', filename)
-        self.queue.put((filename, time.time() + 30))
+        remove_delay = int(self._attrs.get('remove_delay', 1))
+        self.queue.put((filename, time.time() + 30 * remove_delay))
```
pnuu: Would it be better to have the delay directly in the config file without the hidden 30x multiplier?
```python
remove_delay = int(self._attrs.get('remove_delay', 30))
self.queue.put((filename, time.time() + remove_delay)
```

trollmoves/server.py
```diff
@@ -380,19 +382,34 @@ def create_file_notifier(attrs, publisher):
...
-             pyinotify.IN_CREATE)
+             pyinotify.IN_CREATE |
+             pyinotify.IN_DELETE)
 
     wm_ = pyinotify.WatchManager()
 
     pattern = globify(attrs["origin"])
     opath = os.path.dirname(pattern)
 
+    if 'origin_inotify_base_dir_skip_levels' in attrs:
```
pnuu: A short comment here would be nice what's been done.

trollmoves/server.py
```diff
@@ -380,19 +382,34 @@ def create_file_notifier(attrs, publisher):
...
+            elif (os.stat(orig_pathname).st_size == 0):
+                # Want to avoid files with size 0.
+                return
+            else:
+                LOGGER.debug('We have a match: %s', orig_pathname)
         else:
-            LOGGER.debug('We have a match: %s', orig_pathname)
+            # Some systems create temporary file which is then deleted shortly after.
+            # Want to skip these files
+            LOGGER.debug('orig_pathname: {} has disapeared. skipping this.'.format(orig_pathname))
```
pnuu: Is there enough "delay" for the temp file to have been removed? Also, wouldn't the temporary filename _not_ match the configured filename pattern?

trollmoves/client.py
TAlonglong: Yep. Agree

trollmoves/movers.py
TAlonglong: Yes, I did think about something like that, but forgot about it.

So default should be: Do not create remote dir?
This to make it clear to the user if he needs this feature?

trollmoves/movers.py
TAlonglong: You mean from the mkdir command?

I tried to do that, but could not figure out how to catch the output from the exec_command

trollmoves/server.py
TAlonglong: I thought since the original is 30 (seconds that is) that should be default.

But we can use:
```
remove_delay = int(self._attrs.get('remove_delay', 30))
self.queue.put((filename, time.time() + remove_delay))
```

trollmoves/server.py
TAlonglong: OK

trollmoves/movers.py
pnuu: Yes, I think remote dir creation should be turned off by default. Easier to catch typos and such if the feature isn't needed.

trollmoves/movers.py
pnuu: I was thinking there were some "general" exceptions the command could raise. But it seems that the output can be collected straight from the command return values (from http://docs.paramiko.org/en/2.6/api/client.html):
```python
stdin, stdout, stderr = client.exec_command('ls -l')
```
So maybe the `stderr` could be inspected. Can't now figure out how to connect the client so can't check what the `stderr` contains in different cases :smile:

trollmoves/server.py
pnuu: Exactly :-) In the current version in PR the config value would multiply the default delay (30 s) and wouldn't be very intuitive thing to configure.

trollmoves/movers.py
TAlonglong: Hm, looking closely on this, its actually not needed. because there is a `create_local_dir` in the client.py called in `request_push`

trollmoves/server.py
TAlonglong: ok

trollmoves/server.py
TAlonglong: Unfortunately there is enough delay. I have a 3 party system that do that.

But you are right, it should _not_ match. I added this some years ago to my system so I dont remember completely why I added this. I need to check this.

TAlonglong: Here is what I get:
[2019-10-20 06:43:17,034 DEBUG    trollmoves.server] orig_pathname: /data/trollduction/METOP_level0/.IASI_HRP_00_M01_20191020063931Z_20191020064119Z_N_O_20191020063937Z.LXJt3P has disapeared. skipping this.

So if you match like this 
origin=/data/trollduction/METOP_level0/{sensor:s}_HRP_00_{platform_name:3s}_{start_time:%Y%m%d%H%M%S}Z_{end_time:%Y%m%d%H%M%S}Z_N_O_{process_time:%Y%m%d%H%M%S}Z

But anyway it would not match. Maybe I had a to greedy match string when I saw this and added this.

I think it it safe to remove it.

mraspaud:
How are we doing here ? I this ready for final review/merge ?

TAlonglong:
Oh, time flies when you have fun! ( or too much to do :-( )

Think I should add some example as @pnuu suggests. After that ready.

TAlonglong:
OK @mraspaud I think it is ready

PR #46, opened 2019-10-17
Add config option for ssh port. Defaults to 22

TAlonglong:


TAlonglong:
This needs a bit more work as this PR needs the `ssh_port` config to be in the server config, but the mover config is in the client config.

Need to send the port from the client from the server.

TAlonglong:
So I have added the possibility to give the `ssh_port` in the client config.

I see the default sftp port is 22, should this be added to sftp also @pnuu @mraspaud ?

mraspaud:
Yes, it might even be relevant for ftp
One question though: is it not possible or practical to include the port in the destination directly as `ssh://blabla.com:33/my/path` ?

TAlonglong:
Thanks @mraspaud . It is actually better to use it that way you suggest, else it would interfere with ftp default port 21. Also added possible to give port to sftp.

TAlonglong:
Forgot to mention I have not tested with ftp or sftp.

mraspaud:
Do you feel like making some small unit tests for the ssh movers ?

pnuu:
I think it is OK to have the `port` defined as a separate config option. Then it can be given in the `[DEFAULT]` section. And that's the way e.g. `publisher_port` and `request_port` are now configured. The other use case we have that _does_ have the ports in the URI is the *list* of `providers`, where the portnumber needs to be defined that way for it to be definable for each host separately.

TAlonglong:
OK, add tests for the scpmover. 

I think for now the port number should be given with the destination.

PR #55, opened 2019-11-13
Fix dispatcher transfers for scp protocol

pnuu:
This PR fixes a bug that caused all SCP transfers with `dispatcher.py` to be straight file copies.

Also fixes `--disable-backlog` use in `move_it_server`.

### DIFF
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -156,6 +156,8 @@ import yaml
 
 import inotify.adapters
 from inotify.constants import IN_MODIFY, IN_CLOSE_WRITE, IN_CREATE, IN_MOVED_TO
+from six.moves.urllib.parse import urlsplit, urlunsplit
+
 from posttroll.listener import ListenerContainer
 from trollmoves.movers import move_it
 from trollmoves.utils import clean_url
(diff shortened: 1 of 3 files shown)


trollmoves/dispatcher.py
```diff
@@ -330,7 +333,10 @@ def create_dest_url(self, msg, client, item):
             if key in mda:
                 mda[key] = aliases.get(mda[key], mda[key])
         path = compose(path, mda)
-        return urljoin(host, path), connection_parameters
+        parts = urlsplit(host)
+        host_path = urlunsplit((parts.scheme, parts.netloc, path,
+                                parts.query, parts.fragment))
+        return host_path, connection_parameters
```
mraspaud: does the urlunsplit preserve the username and password ? Maybe add a test for that ?

trollmoves/tests/test_dispatcher.py
```diff
@@ -361,6 +361,35 @@ def test_dispatcher():
...
+                config_file.flush()
+                config_file.close()
+            config = yaml.safe_load(test_yaml2)
+            dp = Dispatcher(config_file_name)
+            msg = Mock()
+            msg.subject = '/level2/viirs'
+            msg.data = {'sensor': 'viirs', 'product': 'green_snow', 'platform_name': 'NOAA-20',
+                        'start_time': datetime(2019, 9, 19, 9, 19), 'format': 'tif'}
+            url, params = dp.create_dest_url(msg, 'target2', config['target2'])
+        expected_url = "ssh://server.target2.com/satellite/viirs/sat_201909190919_NOAA-20.tif"
```
mraspaud: Can you also check other schemes like `scp` ?

trollmoves/dispatcher.py
pnuu: It does. Test added.

trollmoves/tests/test_dispatcher.py
pnuu: Test added.

PR #57, opened 2020-03-05
Reload only the actual config file when it is modified

pnuu:
The current behaviour of `move_it.py` causes the script to act on _all_ the changes in the directory where the configuration file is located. That is, if any of the other files in the directory were opened and closed those files were tried to be read as a valid configuration file.

This PR fixes this so that only the _modification_ of the actual configuration file triggers the reload.

The reload by default also triggers resending of all existing files. To prevent this (and transfering possibly a file that is being actively written to) a command line argument `-d` or `--disable-backlog` is added to disable it.

### DIFF
--- a/bin/move_it.py
+++ b/bin/move_it.py
@@ -507,7 +507,12 @@ class EventHandler(pyinotify.ProcessEvent):
         self._fun = fun
 
     def process_IN_CLOSE_WRITE(self, event):
-        """On closing after writing.
+        """On closing a writable file.
+        """
+        self._fun(event.pathname)
+
+    def process_IN_MODIFY(self, event):
+        """The file was modified.
         """
         self._fun(event.pathname)
 
--- a/bin/move_it.py
+++ b/bin/move_it.py
@@ -210,6 +210,7 @@ def reload_config(filename, disable_backlog=False):
 
     old_glob = []
 
+    config_changed = False
     for key, val in new_chains.items():
         identical = True
         if key in chains:
@@ -218,6 +219,7 @@ def reload_config(filename, disable_backlog=False):
                     ((key2 not in chains[key]) or
                      (chains[key][key2] != val2))):
                     identical = False
+                    config_changed = True
                     break
             if identical:
                 continue
@@ -297,7 +299,9 @@ def reload_config(filename, disable_backlog=False):
         del chains[key]
         LOGGER.debug("Removed %s", key)
 
-    LOGGER.debug("Reloaded config from %s", filename)
+    if config_changed:
+        LOGGER.debug("Reloaded config from %s", filename)
+
     if old_glob and not disable_backlog:
         fnames = []
         for pattern in old_glob:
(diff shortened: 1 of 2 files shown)


pnuu:
I'm not exactly sure how the `EventNotifier` gets the arguments, so that needs a verification from @mraspaud .

bin/move_it.py
```diff
@@ -645,13 +653,13 @@ def main():
 
     LOGGER.info("Starting up.")
 
-    mask = (pyinotify.IN_CLOSE_WRITE |
+    mask = (pyinotify.IN_MODIFY |
```
mraspaud: I'm ok with adding in_modify, but I think in_close_write should be there too.

mraspaud: Actually I'm not sure anymore. I think we should wait until the file is closed before we reload it. Otherwise we take the risk of reloading a partially modified file if the editor of the user if saving automatically for example.

pnuu: `IN_CLOSE_WRITE` means that the config will be reloaded everytime the file is opened for viewing and then closed. It literally means "writable file was closed". `IN_MODIFY` means the file was modified, and that only happens when it is saved. Sure, automatical saving can cause issues, but it's either that, reload even after `cat file.yaml`, or no automatic updates. I vote for the implementation I have here.

mraspaud: `cat` doesn't open the file in write mode. I think `IN_CLOSE_WRITE` is more appropriate since it won't trigger a reload with a half finished file that is saved. And the `reload_config` function should in principle not restart anything if the file hasn't changed.

pnuu: After some thought, I reverted this event to `IN_CLOSE_WRITE` and adjusted the logging so that it doesn't seem like something was changed if the file was just opened and closed with an editor.

bin/move_it.py
```diff
@@ -297,8 +299,10 @@ def delete_hook(pathname, val=val, pub=pub):
         del chains[key]
         LOGGER.debug("Removed %s", key)
 
-    LOGGER.debug("Reloaded config from %s", filename)
-    if old_glob:
+    if config_changed:
+        LOGGER.debug("Reloaded config from %s", filename)
```
mraspaud:
```suggestion
    if config_changed:
        LOGGER.debug("Reloaded config from %s", filename)
    else:
        LOGGER.debug("No changes to reload in %s", filename)

```

bin/move_it.py
```diff
@@ -507,7 +512,12 @@ def __init__(self, fun, *args, **kwargs):
         self._fun = fun
 
     def process_IN_CLOSE_WRITE(self, event):
-        """On closing after writing.
+        """On closing a writable file.
+        """
+        self._fun(event.pathname)
+
+    def process_IN_MODIFY(self, event):
+        """The file was modified.
```
mraspaud: This is a change in behaviour from previously, where nothing would happen until the file was closed. Maybe it should be documented ? (in the module docstring if since we don't have proper docs)

pnuu: So `EventHandler` acts on these events even if it's not in the `mask` that is set up in `main()`? Not that this is used there, so might as well remove it.

pnuu: The `process_IN_MODIFY()` is now removed.

mraspaud: Sorry, missed the mask! but thanks for removing it.

PR #58, opened 2020-03-20
Publish messages after dispatch

pnuu:
This PR adds a possibility to publish a message after the file has been transferred using `dispatcher.py`.

### DIFF
--- a/bin/dispatcher.py
+++ b/bin/dispatcher.py
@@ -86,7 +86,8 @@ def main():
                         "verbose logging). WARN=0, INFO=1, "
                         "DEBUG=2. This is overridden by the log config file if specified.")
     parser.add_argument(
-        "-p", "--publish-port", type=int, dest="pub_port",
+        "-p", "--publish-port", type=int, dest="pub_port", nargs='?',
+        const=0, default=None,
         help="Publish messages for dispatched files on this port. "
         "Default: no publishing.")
     parser.add_argument("-n", "--publish-nameserver", nargs='*',
--- a/bin/dispatcher.py
+++ b/bin/dispatcher.py
@@ -85,13 +85,21 @@ def main():
                         help="Verbosity (between 1 and 2 occurrences with more leading to more "
                         "verbose logging). WARN=0, INFO=1, "
                         "DEBUG=2. This is overridden by the log config file if specified.")
+    parser.add_argument(
+        "-p", "--publish-port", type=int, dest="pub_port",
+        help="Publish messages for dispatched files on this port. "
+        "Default: no publishing.")
+    parser.add_argument("-n", "--publish-nameserver", nargs='*',
+                        dest="pub_nameservers",
+                        help="Nameserver for publisher to connect to")
     cmd_args = parser.parse_args()
-
     setup_logging(cmd_args)
     logger.info("Starting up.")
 
     try:
-        dispatcher = Dispatcher(cmd_args.config_file)
+        dispatcher = Dispatcher(cmd_args.config_file,
+                                publish_port=cmd_args.pub_port,
+                                publish_nameservers=cmd_args.pub_nameservers)
     except Exception as err:
         logger.error('Dispatcher crashed: %s', str(err))
         sys.exit(1)
@@ -102,6 +110,7 @@ def main():
         logger.debug("Interrupting")
     finally:
         dispatcher.close()
+
 
 if __name__ == '__main__':
     main()
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -48,6 +48,10 @@ Example config::
       # subscribe_services:
       #   - service_name_1
       #   - service_name_2
+      # Message topics for published messages. Required if command-line option
+      #   "-p"/"--publish-port" is used.  The topic can be composed using
+      #   metadata from incoming message
+      # publish_topic: "/new/topic/{platform_name}"
       aliases:
         product:
           natural_color: dnc
@@ -330,6 +334,7 @@ class Dispatcher(Thread):
             info = msg.data.copy()
             info["uri"] = urlsplit(url).path
             topic = self.config[client].get("publish_topic", msg.subject)
+            topic = compose(topic, info)
             msg = Message(topic, 'file', info)
             logger.debug('Publishing %s', str(msg))
             self.publisher.send(str(msg))
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -268,10 +268,11 @@ class Dispatcher(Thread):
         self.topics = None
         self.listener = None
         self.publisher = None
-        if publish_port:
+        if publish_port is not None:
             self.publisher = NoisyPublisher(
                 "dispatcher", port=publish_port,
                 nameservers=publish_nameservers)
+            self.publisher.start()
         self.loop = True
         self.config_handler = DispatchConfig(config_file, self.update_config)
         signal.signal(signal.SIGTERM, self.signal_shutdown)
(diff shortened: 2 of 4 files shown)


pnuu:
Still needs couple of unit tests. I'll add them on monday.

bin/dispatcher.py
```diff
@@ -85,13 +85,21 @@ def main():
                         help="Verbosity (between 1 and 2 occurrences with more leading to more "
                         "verbose logging). WARN=0, INFO=1, "
                         "DEBUG=2. This is overridden by the log config file if specified.")
+    parser.add_argument(
+        "-p", "--publish-port", type=int, dest="pub_port",
+        help="Publish messages for dispatched files on this port. "
+        "Default: no publishing.")
+    parser.add_argument("-n", "--publish-nameserver", nargs='*',
+                        dest="pub_nameservers",
+                        help="Nameserver for publisher to connect to")
```
mraspaud: This defaults to None, right ?

bin/dispatcher.py
```diff
@@ -85,13 +85,21 @@ def main():
                         help="Verbosity (between 1 and 2 occurrences with more leading to more "
                         "verbose logging). WARN=0, INFO=1, "
                         "DEBUG=2. This is overridden by the log config file if specified.")
+    parser.add_argument(
+        "-p", "--publish-port", type=int, dest="pub_port",
+        help="Publish messages for dispatched files on this port. "
+        "Default: no publishing.")
```
mraspaud: Can you set this to 0 if -p is provided but not number is given ? (random port)

trollmoves/dispatcher.py
```diff
@@ -434,15 +474,20 @@ def dispatch(source, destinations):
...
         logger.error(message)
         any_error = True
+    success = {}
     # rename and send file with right protocol
-    for url, params in destinations:
+    for url, params, client in destinations:
         try:
             logger.debug("Dispatching %s to %s", source, str(clean_url(url)))
             move_it(source, url, params)
+            success[client] = True
```
mraspaud: I thinks bad things are happening if we have multiple destinations on the same client

mraspaud: This should be tested

trollmoves/dispatcher.py
```diff
@@ -336,13 +374,15 @@ def create_dest_url(self, msg, client, item):
...
+        return host_path, connection_parameters, client
 
     def close(self):
         """Shutdown the dispatcher."""
         logger.info('Terminating dispatcher.')
         self.loop = False
         self.listener.stop()
+        if self.publisher:
+            self.publisher.stop()
         self.config_handler.close()
```
mraspaud: Can we put each of these in  a try/except to make sure the following are executed too ?

mraspaud: Maybe with a logging message giving the reason of the failure.

trollmoves/tests/test_dispatcher.py
```diff
@@ -397,17 +399,69 @@ def test_create_dest_url():
...
+        dp.publisher.send.assert_called_once()
+        # The message topic has been composed and uri has been replaced
+        msg_call = call('/topic/platform', 'file',
+                        {'uri': 'url2', 'platform_name': 'platform'})
+        assert msg_call in Message.mock_calls
+    finally:
+        os.remove(config_file_name)
+        if dp is not None:
+            dp.close()
+            dp.publisher.stop.assert_called()
```
mraspaud: Why not push all this one step in and let the NamedTemporaryFile delete the file ?

trollmoves/dispatcher.py
```diff
@@ -241,30 +247,41 @@ class DispatchConfig(YAMLConfig):
...
         super().read_config()
         self.callback(self.config)
 
 
 class Dispatcher(Thread):
     """Class that dispatches files."""
 
-    def __init__(self, config_file):
+    def __init__(self, config_file, publish_port=None,
+                 publish_nameservers=None):
```
mraspaud: Could you test that not providing the nameservers on the command line actually gives None, and in that case the regular multicast system is used?

bin/dispatcher.py
pnuu: Yes.

bin/dispatcher.py
pnuu: The option requires an argument, so plain `-p` won't work.

mraspaud: I think there is a parameter for this in argparse.

trollmoves/dispatcher.py
pnuu: Oh, that's possible? Ok, will have to think about it a bit.

trollmoves/dispatcher.py
pnuu: Will do.

trollmoves/tests/test_dispatcher.py
pnuu: Done.

trollmoves/dispatcher.py
pnuu: Done.

bin/dispatcher.py
pnuu: Updated. Calling `dispatcher.py config.yaml -p` will result in
`Namespace(config_file='config.yaml', log=None, log_config=None, pub_nameservers=None, pub_port=0, verbosity=0)`.

Calling `dispatcher.py config.yaml` will result in
`Namespace(config_file='config.yaml', log=None, log_config=None, pub_nameservers=None, pub_port=None, verbosity=0)`

Is this what you suggested?

mraspaud: yes, perfect, thanks !

trollmoves/dispatcher.py
pnuu: As discussed on Slack, I added an exception here in the case there were multiple clients with the same name. And added few tests for `dispatch()`.

trollmoves/dispatcher.py
```diff
@@ -299,7 +317,27 @@ def run(self):
...
+        The URI is replaced with the URI on the target server.
+
+        """
+        for url, params, client in destinations:
+            if not success[client]:
+                continue
+            del params
+            info = msg.data.copy()
+            info["uri"] = urlsplit(url).path
+            topic = self.config[client].get("publish_topic", msg.subject)
```
mraspaud: The example says the publish topic is required, but here you have a default ?

trollmoves/tests/test_dispatcher.py
```diff
@@ -397,17 +399,124 @@ def test_create_dest_url():
...
+"""
+@patch('trollmoves.dispatcher.Message')
+@patch('trollmoves.dispatcher.ListenerContainer')
+@patch('trollmoves.dispatcher.NoisyPublisher')
+def test_publisher(NoisyPublisher, ListenerContainer, Message):
+    """Test the publisher is initialized."""
+    pub = Mock()
+    NoisyPublisher.return_value = pub
+    try:
+        with NamedTemporaryFile('w', delete=False) as config_file:
```
mraspaud: Why not letting `delete=True` ?

trollmoves/dispatcher.py
pnuu: Ah, you're right. I'll update the example.

trollmoves/tests/test_dispatcher.py
pnuu: Because the file would be deleted when _closed_, not when the context manager exits. So when the file is written to and closed, the context manager will delete it and the config can't be read.

mraspaud: Ah, good point! that's an important detail from the documentation :)

trollmoves/dispatcher.py
pnuu: Actually I went the other way around, log an error when the topic is missing.

mraspaud: Sounds good, I could create a feedback look if the original topic is used.

PR #59, opened 2020-03-23
Fix rpm names for newer distros

mraspaud:
This fixes trollmoves to require the right packages with rhel8 among others.


mraspaud:
@pnuu @TAlonglong any opinion on this ?

mraspaud:
@TAlonglong this means that if this module is build as rpm package (centos, rhel, ...) and installed, it will use python3

pnuu:
LGTM, although I have no experience in building RPMs.

PR #60, opened 2020-03-30
Fix iterate_messages() generator raising StopIteration

pnuu:
This PR fixes `iterate_messages()` what was raising `StopIteration` when it should just return.

Also adds an unit test for this function, and fixes a flake8 complaint about invalid usage of `assert`:

`trollmoves/tests/test_client.py:86:9: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().`

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -471,7 +471,7 @@ def iterate_messages(uid):
         try:
             msg = get_next_msg(uid)
         except IndexError:
-            raise StopIteration
+            return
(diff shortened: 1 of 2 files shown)


mraspaud:
For the record, this is new behavious since python 3.7:
> Changed in version 3.7: Enable PEP 479 for all code by default: a StopIteration error raised in a generator is transformed into a RuntimeError.

PR #62, opened 2020-04-28
Add Watchdog based event handler

pnuu:
This PR adds an event handler based on `watchdog` library. It can be used in cases where `inotify` messages are not emitted by the OS, like NFS and SSHFS mounts.

The `watchdog` behaviour can be turned on by defining `-w` or `--watchdog` on the command line when starting `move_it_server.py`. Otherwise the usage and configuration are identical to when using inotify.

I'll have a look if there are tests that could be added. Currently the "server tests" just seem to test the `trollmoves.utils` module.


pnuu:
Seems that `watchdog` only reports the latest file that has arrived within the defined `timeout`. Shortening the `timeout` from the default 1.0 seconds to 0.1 seconds increases the CPU load from ~9 % to ~26 %. To start with, that's a lot.

pnuu:
Nevermind, I misread the log output on terminal, all the new files are reported. And Eumetcast Terrestrial worked as a new source when mounted via sshfs and `-w` switch was used.
I also fixed the `--disable-backlog` usage, it wasn't working when the configuration file was reloaded after hot-modifying.

trollmoves/server.py
```diff
@@ -702,29 +767,40 @@ def process_IN_DELETE(self, event):
...
+def process_old_files(pattern, fun, publisher, kwargs):
+    """Process files from *pattern* with function *fun*."""
     fnames = glob.glob(pattern)
     if fnames:
         # time.sleep(3)
         LOGGER.debug("Touching old files")
         for fname in fnames:
             if os.path.exists(fname):
-                fun(fname)
+                fun(fname, publisher, pattern, kwargs)
```
mraspaud, line 792: `create_inotify_notifier` return a function that only takes a filename, how can this work ?

trollmoves/move_it_base.py
```diff
@@ -53,26 +55,32 @@ def __init__(self, cmd_args, chain_type):
         self.setup_watchers(cmd_args)
 
     def reload_cfg_file(self, filename, *args, **kwargs):
+        """Reload configuration file."""
         if self.chain_type == "client":
             from trollmoves.client import reload_config
-            reload_config(filename, self.chains, *args, pub_instance=self.pub,
-                          **kwargs)
```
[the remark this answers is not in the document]
pnuu: Seems to be an accidental deletion, will add it back.

trollmoves/move_it_base.py
```diff
@@ -53,26 +55,32 @@ def __init__(self, cmd_args, chain_type):
...
         else:
             # Also Mirror uses the reload_config from the Server
             from trollmoves.server import reload_config
             reload_config(filename, self.chains, *args, publisher=self.pub,
-                          **kwargs)
```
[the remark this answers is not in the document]
pnuu: It seems that they are never set to anything in the calls we have in Trollmoves and the default kwarg values are used. In `MoveItBase` the `self.notifier` and `self.watchman` only handle the configuration file, where as `trollmoves.server.reload_config()` and `trollmoves.client.reload_config()` handle the incoming files. I'd say the `manager` and `notifier_builder` kwargs are useless.

trollmoves/server.py
```diff
@@ -365,21 +380,50 @@ def run(self):
...
     listener = Listener(attrs, publisher)
 
     return listener, None
 
 
-def create_file_notifier(attrs, publisher):
-    """Create a notifier from the specified configuration attributes *attrs*.
-    """
-
+def process_notify(orig_pathname, publisher, pattern, kwargs):
```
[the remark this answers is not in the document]
pnuu: Done.

trollmoves/server.py
pnuu: The rest of the arguments were forgotten from the `fun()` definition, the call it was doing had them.

pnuu:
Oh, and yes, I agree, the code is pretty messy. The refactoring I did a year(?) ago helped a bit, but the confusing notifiers and managers and returned functions and non-descriptive naming (managers and notifiers in many places with different uses) makes the code hard to follow. Trollmoves would need some work. Maybe not a full rewrite, but at least the attribute naming should be clarified.

trollmoves/server.py
```diff
@@ -450,7 +450,7 @@ def create_inotify_notifier(attrs, publisher):
         opath = os.path.join("/", pattern_join)
         LOGGER.debug("Using %s as base path for pyinotify add_watch.", opath)
 
-    def fun(orig_pathname):
+    def fun(orig_pathname, publisher, pattern, attrs):
         """Wrap the publisher."""
         process_notify(orig_pathname, publisher, pattern, attrs)
```
mraspaud: why not return `process_notify` directly then ?

pnuu: Because I followed the existing structure without thinking too much :roll_eyes: 

Simplified and removed the nested function.

PR #67, opened 2020-05-25
Accept uri's with a full scheme (ssh, ftp,...)

adybbroe:
Currently trollmove's dispatcher expects a file-path as uri, and does not handle a full url including scheme (ssh, sftp, etc). This PR attempts to solve that.

### DIFF
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2012-2019
+# Copyright (c) 2012-2020
 #
 # Author(s):
 #
@@ -160,7 +160,8 @@ import yaml
 
 import inotify.adapters
 from inotify.constants import IN_MODIFY, IN_CLOSE_WRITE, IN_CREATE, IN_MOVED_TO
-from six.moves.urllib.parse import urlsplit, urlunsplit
+from six.moves.urllib.parse import urlsplit, urlunsplit, urlparse
+import socket
 
 from posttroll.listener import ListenerContainer
 from posttroll.publisher import NoisyPublisher
@@ -268,6 +269,8 @@ class Dispatcher(Thread):
         self.topics = None
         self.listener = None
         self.publisher = None
+        self.host = socket.gethostname()
+
         if publish_port is not None:
             self.publisher = NoisyPublisher(
                 "dispatcher", port=publish_port,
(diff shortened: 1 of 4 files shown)


trollmoves/dispatcher.py
```diff
@@ -318,7 +321,19 @@ def run(self):
                     continue
                 destinations = self.get_destinations(msg)
                 if destinations:
-                    success = dispatch(msg.data['uri'], destinations)
+                    # Check if the url are on another host:
+                    url = urlparse(msg.data['uri'])
+                    if url.scheme != '':
+                        netloc = url.netloc
```
pnuu, line 327: This is unnecessary

trollmoves/dispatcher.py
```diff
@@ -318,7 +321,19 @@ def run(self):
...
+                        if netloc != self.host:
+                            # This warning may appear even if the file path
+                            # does exist on both servers (a central file server
+                            # reachable from both). But in those cases one
+                            # should probably not use an url scheme but just a
+                            # file path:
+                            logger.warning(
+                                "uri is pointing to a file path on another server! Host=<%s> uri netloc=<%s>", self.host, netloc)
```
pnuu, line 335: Please shorten/wrap this line.

trollmoves/dispatcher.py
```diff
@@ -318,7 +321,19 @@ def run(self):
                     continue
                 destinations = self.get_destinations(msg)
                 if destinations:
-                    success = dispatch(msg.data['uri'], destinations)
+                    # Check if the url are on another host:
+                    url = urlparse(msg.data['uri'])
+                    if url.scheme != '':
+                        netloc = url.netloc
+                        if netloc != self.host:
```
mraspaud, line 328: This doesn't always work. See here how it is solved:
https://github.com/pytroll/trollmoves/blob/master/trollmoves/server.py#L351-L352

Maybe a function could be made for this and used in both places ?

trollmoves/dispatcher.py
```diff
@@ -318,7 +321,19 @@ def run(self):
...
+                            # This warning may appear even if the file path
+                            # does exist on both servers (a central file server
+                            # reachable from both). But in those cases one
+                            # should probably not use an url scheme but just a
+                            # file path:
+                            logger.warning(
```
mraspaud, line 334: Should this raise a `NotImplementedError` ?

adybbroe:
> Couple of small comments. And please add a description :-)

Done! :-)

trollmoves/dispatcher.py
adybbroe: Well, probably/strictly you are right, but I don't like this:
```
In [7]: path2 = '//san1/polar_out/direct_readout/viirs_active_fires/filtered/AFIMG_j01_d20200423_t110200.geojson'                                                      

In [8]: url2 = urlparse(path2)                                                                                                                                         

In [10]: url2.netloc                                                                                                                                                   
Out[10]: 'san1'
```
In theory `san1` could be the name of a host?

trollmoves/dispatcher.py
adybbroe: Ok, will do.

trollmoves/dispatcher.py
adybbroe: Yes, good idea!
Done

mraspaud:
Tests are failing, I think you should use `from trollmoves.utils ...`

adybbroe:
> Tests are failing, I think you should use `from trollmoves.utils ...`

Oh sorry!

adybbroe:
Great thanks!

PR #71, opened 2020-05-26
Fix subscriber stopping

mraspaud:
Refactor client's reload_config

- [x] Add tests for Chain
- [x] Fix subscriber


examples/move_it_client.ini
```diff
@@ -16,6 +16,10 @@ topic = /1b/hrit-segment/0deg
 publish_port = 0
 # Disable multicasting and use direct nameserver connections
 nameservers = localhost 192.168.0.10 192.168.0.11
+# Activate heartbeat monitoring
+# Allow missing 10 hearbeats (every 30 seconds):
```
TAlonglong, line 20: just a minor, minor comment: every 30 seconds should maybe be: 'every heartbeat last 30 seconds(default)'

mraspaud, line 20:
```suggestion
# Allow missing 10 hearbeats (every heartbeat last 30 seconds(default)):
```

PR #73, opened 2020-06-02
Make filepattern optional

### LINKED ISSUE #72: filepattern is a required parameter - should be optional
I want to use the diapatcher to dispatch files to our archive. I want to be absolutely sure the file names are kept as is, so I do not want to specify a filepattern to my yaml file. However, not adding a filepattern breaks my process:
https://github.com/pytroll/trollmoves/blob/041ae4feb3c6f7a0dd96f5942a3a63d8d195e1d7/trollmoves/dispatcher.py#L387

pnuu:
This PR makes `filepattern` an optional config item. If not available, the target filename will be taken from the incoming message `msg.data['uid']`, which is assumed to be available.

Closes #72

### DIFF
--- a/examples/dispatch.yaml
+++ b/examples/dispatch.yaml
@@ -2,6 +2,8 @@ target1:
   host: ftp://ftp.target1.com
   connection_parameters:
     connection_uptime: 60
+  # File name pattern for the target file.  If not given, the filename
+  #   will be taken from the incoming message 'uid'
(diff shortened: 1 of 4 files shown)


mraspaud:
@adybbroe there are unittests implemented for this in this PR, or are there others you need?

pnuu:
I agree with the changes @mraspaud made, so merging this and making a new release.

PR #78, opened 2020-06-05
Get username and password from .netrc file if available

adybbroe:
Check if username and password are available in the .netrc file for the host of the destination

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -34,10 +34,10 @@ from threading import Event, Lock, Thread, current_thread
 import netrc
 
 from six import string_types
-from six.moves.urllib.parse import urlparse
+#from six.moves.urllib.parse import urlparse
+from furl import furl as urlparse
 
 from trollmoves.utils import clean_url
-
 from paramiko import SSHClient, SSHException, AutoAddPolicy
 from scp import SCPClient
 
@@ -60,8 +60,10 @@ def move_it(pathname, destination, attrs=None, hook=None, rel_path=None):
     new_dest = dest_url._replace(path=new_path)
     fake_dest = clean_url(new_dest)
 
+    LOGGER.debug("new_dest = %s", new_dest)
     LOGGER.debug("Copying to: %s", fake_dest)
     try:
+        LOGGER.debug("Scheme = %s", str(dest_url.scheme))
         mover = MOVERS[dest_url.scheme]
     except KeyError:
         LOGGER.error("Unsupported protocol '" + str(dest_url.scheme)
@@ -92,6 +94,10 @@ class Mover(object):
         else:
             self.destination = destination
 
+        self.dest_username = None
+        self.dest_password = None
+
+        LOGGER.debug("Destination: %s", str(destination))
         self.origin = origin
         self.attrs = attrs or {}
 
@@ -108,6 +114,8 @@ class Mover(object):
     def get_connection(self, hostname, port, username=None):
         """Get the connection."""
         with self.active_connection_lock:
+            LOGGER.debug("Destination username and passwd: %s %s",
+                         self.destination.username, self.destination.password)
             LOGGER.debug('Getting connection to %s@%s:%s',
                          self.destination.username, self.destination.hostname, port)
             try:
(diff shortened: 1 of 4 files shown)


trollmoves/movers.py
```diff
@@ -42,6 +43,12 @@
 
 LOGGER = logging.getLogger(__name__)
 
+HOME = os.environ.get('HOME')
+if HOME:
+    NETRCFILE = os.path.join(HOME, '.netrc')
+else:
+    NETRCFILE = None
+
```
mraspaud, line 51: This isn't needed. According to the documentation of netrc, the filename is optional and defaults to home's .netrc: https://docs.python.org/3.8/library/netrc.html#netrc.netrc

trollmoves/movers.py
```diff
@@ -201,10 +209,30 @@ class FtpMover(Mover):
...
+                    self.destination.hostname)
+                logger.debug('Got username and password from netrc file!')
+
     def open_connection(self):
+        """Open the connection and login."""
         connection = FTP(timeout=10)
         connection.connect(self.destination.hostname,
                            self.destination.port or 21)
+
+        self._get_netrc_authentication()
```
mraspaud, line 235: This should be moved a couple of lines further down, so as not to overwrite username and password if they were provided in `destination`

trollmoves/movers.py
adybbroe: Well, you are right. However, this came from experience with another application where HOME is not defined. As you know when run from supervisor HOME is not available. Therefore I implemented this. If home is not set you will get the following in your case:
```
In [5]: secrets = netrc.netrc()                                                                                                                                               
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-5-211a7aaf8c74> in <module>
----> 1 secrets = netrc.netrc()

~home/a000680/miniconda3/envs/python38/lib/python3.8/netrc.py in __init__(self, file)
     27         self.hosts = {}
     28         self.macros = {}
---> 29         with open(file) as fp:
     30             self._parse(file, fp, default_netrc)
     31 

FileNotFoundError: [Errno 2] No such file or directory: '/.netrc'
```
@mraspaud Do you suggest I catch that exception then and log a warning?

mraspaud: if `$HOME` isn't available, the `netrc` module will surely raise an exception, why not just catch it ?

mraspaud: So the answer to your question is 'yes'.

mraspaud: also note that is the `.netrc` file isn't present in `$HOME`, you'll have to catch a `FileNotFoundError`

trollmoves/movers.py
adybbroe: Ok. Will do. Please note that _get_netrc_authentication only change the username and password if the netrc file is available and the hostname is included there.

trollmoves/movers.py
adybbroe: @mraspaud Can you please give me a hint where you would have such a small test? I find no tests for movers.py, so I suppose it somewhere in test_dispatcher.py? These tests seem to test on a higher level, or?

mraspaud: I didn't realise there wasn't any test for movers. Could you create a test file for movers and place just a simple test there for the ftpmover's change you made ? If the file is there, we can then populate it with more tests

adybbroe:
I pushed something accidentally. I was testing furl, as with urlparse I can't change username or password:
```
In [1]: from six.moves.urllib.parse import urlparse                                                                                                                           
In [2]: destination = 'ftp://lxserv1096.smhi.se/san1/test_msb_dispatch/afimg_20200604_125613.geojson'                                                                                                                                                                 
In [3]: dest_url = urlparse(destination)
In [4]: dest_url.username                                                                                                                                 
In [5]: dest_url.username = 'xxx'                                                                                                                                             
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-588222ecef73> in <module>
----> 1 dest_url.username = 'xxx'

AttributeError: can't set attribute
```

trollmoves/tests/test_movers.py
```diff
@@ -0,0 +1,68 @@
...
+    username = 'myusername'
+    password = 'mypasswd'
+    account = None
+
+    origin = '/path/to/mydata/filename.ext'
+
+    def raise_my_error():
+        raise FileNotFoundError
+
+    netrc.side_effect = raise_my_error
```
pnuu, line 43: The side effect can be directly set to `FileNotFoundError`, no need to wrap it in a function.

trollmoves/movers.py
```diff
@@ -201,13 +211,38 @@ class FtpMover(Mover):
...
+        except (netrc.NetrcParseError, FileNotFoundError) as e__:
+            LOGGER.warning('Failed retrieve authentification details from netrc file! Exception: ', str(e__))
+            return
+
+        LOGGER.debug("hosts: %s", str(list(secrets.hosts.keys())))
+        LOGGER.debug("Check if hostname matches any listed in the netrc file")
+        if self.destination.hostname in list(secrets.hosts.keys()):
+            LOGGER.debug("Destination hostname: %s", self.destination.hostname)
+            self._dest_username, account, self._dest_password = secrets.authenticators(self.destination.hostname)
+            LOGGER.debug("Destination hostname: %s", self.destination.hostname)
```
pnuu, line 228: This log message is done twice.

trollmoves/movers.py
```diff
@@ -201,13 +211,38 @@ class FtpMover(Mover):
...
+    def _get_netrc_authentication(self):
+        """Get login authentications from netrc file if available"""
+        LOGGER.debug("HOME = %s", os.environ.get('HOME'))
+        try:
+            secrets = netrc.netrc()
+        except (netrc.NetrcParseError, FileNotFoundError) as e__:
+            LOGGER.warning('Failed retrieve authentification details from netrc file! Exception: ', str(e__))
+            return
+
+        LOGGER.debug("hosts: %s", str(list(secrets.hosts.keys())))
```
pnuu, line 223: Is this log message necessary?

trollmoves/movers.py
```diff
@@ -201,13 +211,38 @@ class FtpMover(Mover):
     active_connections = dict()
     active_connection_lock = Lock()
 
+    def _get_netrc_authentication(self):
+        """Get login authentications from netrc file if available"""
+        LOGGER.debug("HOME = %s", os.environ.get('HOME'))
```
pnuu, line 216: Is this log message necessary?

trollmoves/movers.py
adybbroe: Oh, yes, it is probably quite verbose now, as I needed some indications on where it crached until I found the error... I will reduce!

trollmoves/movers.py
adybbroe: Well, hopefully nut, I just needed extra debugging info during development

trollmoves/tests/test_movers.py
adybbroe: Hmm, this was the way I got it working, spent some time on this already :-(
This doesn't work:
```
E       netrc.side_effect = raise FileNotFoundError
E                           ^
E   SyntaxError: invalid syntax
```
How do you mean @pnuu ?

mraspaud: just do
```
netrc.side_effect = FileNotFoundError
```
should work in principle

PR #82, opened 2020-07-27
Try to reply to invalid messages

### LINKED ISSUE #81: Not answering an invalid request could lock the client
https://github.com/pytroll/trollmoves/blob/164410247d7be1622fbc1ac645b421aaa336c7e4/trollmoves/server.py#L294-L296

Ideally we should respond with an error.

mraspaud:
Upon receiving an invalid message, the server is now replying if possible. 
Fixes #81

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -296,7 +296,7 @@ class RequestManager(Thread):
                     try:
                         address = multiparts[0]
                     except (TypeError, IndexError):
-                        LOGGER.warning("Address unknow, not sending an error message back.")
+                        LOGGER.warning("Address unknown, not sending an error message back.")
                     else:
                         message = Message('error', 'error', "Invalid message received")
                         Thread(target=self.reply_and_send,
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -53,7 +53,7 @@ from six.moves.urllib.parse import urlparse
 from trollmoves.client import DEFAULT_REQ_TIMEOUT
 from trollmoves.movers import move_it
 from trollmoves.utils import (clean_url, gen_dict_contains, gen_dict_extract,
-                              get_local_ips, is_file_local)
+                              is_file_local)
 from trollsift import globify, parse
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


adybbroe:
@mraspaud Where are we with this one?

mraspaud:
waiting for reviews

trollmoves/server.py
```diff
@@ -292,7 +292,16 @@ def _run(self):
...
                     address, _, payload = multiparts
                 except ValueError:
-                    LOGGER.warning("Invalid request, ignoring.")
+                    LOGGER.warning("Invalid request.")
+                    try:
+                        address = multiparts[0]
+                    except (TypeError, IndexError):
+                        LOGGER.warning("Address unknow, not sending an error message back.")
+                    else:
+                        message = Message('error', 'error', "Invalid message recieved")
```
pnuu, line 301: Typo, sholud be `received`

trollmoves/server.py
```diff
@@ -292,7 +292,16 @@ def _run(self):
                 try:
                     address, _, payload = multiparts
                 except ValueError:
-                    LOGGER.warning("Invalid request, ignoring.")
+                    LOGGER.warning("Invalid request.")
+                    try:
+                        address = multiparts[0]
+                    except (TypeError, IndexError):
+                        LOGGER.warning("Address unknow, not sending an error message back.")
```
pnuu, line 299: Typo, should be `unknown`

trollmoves/server.py
```diff
@@ -292,7 +292,16 @@ def _run(self):
...
+                    else:
+                        message = Message('error', 'error', "Invalid message recieved")
+                        Thread(target=self.reply_and_send,
+                               args=(self.unknown, address, message)).start()
```
pnuu, line 303: Does this thread terminate by itself?

pnuu: `self.unknown` isn't defined in the `RequestManager` class. Is it something inherited from the `Thread` class?

pnuu (commented):
Couple typos and questions, otherwise LGTM.

trollmoves/server.py
mraspaud:
```suggestion
                        message = Message('error', 'error', "Invalid message received")
```

trollmoves/server.py
mraspaud:
```suggestion
                        LOGGER.warning("Address unknown, not sending an error message back.")
```

trollmoves/server.py
pnuu: Nevermind, `self.unknown` is already there in the current `main` branch. But why does this show here in the diff as an addition?

pnuu (approved):

trollmoves/server.py
mraspaud: yes, the thread sends the answer and dies

PR #83, opened 2020-08-06
Add configurable poll interval (timeout) to watchdog poller

pnuu:
This PR adds a new configuration option for `move_it_server` to lower the polling interval when using `watchdog` notifier (via `-w` command line option). This can lower the CPU usage significantly especially when the monitored directory is mounted over `sshfs`.


pnuu:
Hmm. I had `station` in my old configs, so assumed it is something that is used e.g. in messages or something, but apparently not. I'll adjust the example config and use something else as an example in `[DEFAULT]`.

PR #84, opened 2020-11-27
Fix publisher handling

pnuu:
This PR fixes publisher initialization in `move_it_server.py` when the  configuration file is reloaded.

### DIFF
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -54,9 +54,7 @@ class MoveItMirror(MoveItBase):
 
     def signal_reload_cfg_file(self, *args):
         del args
-        reload_config(self.cmd_args.config_file, self.chains,
-                      self.create_listener_notifier,
-                      MirrorRequestManager, self.sync_publisher)
+        self.reload_cfg_file(self.cmd_args.config_file)
 
     def create_listener_notifier(self, attrs, publisher):
 
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -20,56 +20,61 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Move it Mirror."""
 
-import logging
 import logging.handlers
-import os
 from threading import Lock, Timer
-from six.moves.urllib.parse import urlparse, urlunparse
+from urllib.parse import urlunparse
 import argparse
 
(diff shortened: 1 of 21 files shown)


bin/move_it_mirror.py
```diff
@@ -43,22 +43,22 @@ class MoveItMirror(MoveItBase):
 
     def __init__(self, cmd_args):
         """Set up the mirror."""
-        super(MoveItMirror, self).__init__(cmd_args, "mirror")
-        self.cache_lock = Lock()
         LOGGER.info("Starting publisher on port %s.", str(cmd_args.port))
-        self.pub = Publisher("tcp://*:" + str(cmd_args.port), "move_it_mirror")
+        publisher = Publisher("tcp://*:" + str(cmd_args.port), "move_it_mirror")
+        super(MoveItMirror, self).__init__(cmd_args, "mirror", publisher=publisher)
```
mraspaud, line 48: This looks nearly identical to the server version, can it be factorize into the base class? or a new intermediate class?

bin/move_it_mirror.py
```diff
@@ -43,22 +43,22 @@ class MoveItMirror(MoveItBase):
...
+        self.cache_lock = Lock()
 
     def reload_cfg_file(self, filename):
         """Reload the config file."""
         reload_config(filename, self.chains, self.create_listener_notifier,
-                      MirrorRequestManager, self.pub)
+                      MirrorRequestManager, publisher=self.sync_publisher)
```
mraspaud, line 54: This function and the next are the same, one should just call the other

mraspaud (changes requested):
I think the only test needed here is one checking that the publisher is created for mirror and server. Other than this, I just have a couple of styling suggestions.

bin/move_it_mirror.py
pnuu: I moved the creation to a function defined in `trollmoves.move_it_base`. It can't be created within the base class as we need to recycle it when config is reloaded.

bin/move_it_mirror.py
pnuu: Done.

pnuu:
There are no tests for `trollmoves.move_it_base` so I added the test for creating the publisher for Server to `test_server.py`. I'll move it o `test_base.py` so we can later start adding tests for the base if needed.

mraspaud (approved):
LGTM

PR #94, opened 2021-06-21
Fix attribute passing to the mirror deleter

mraspaud:
The mirroring script was initialised wrongly as not information about the chain was passed on to the deleter.

This PR fixes this.

The problem originated in #20

### DIFF
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -22,10 +22,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Move it Mirror."""
 
-import logging
 import logging.handlers
 from threading import Lock, Timer
-from six.moves.urllib.parse import urlunparse
+from urllib.parse import urlunparse
 import argparse
 
 from posttroll.message import Message
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -20,59 +20,64 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Move it Mirror."""
 
 import logging
 import logging.handlers
-import os
 from threading import Lock, Timer
-from six.moves.urllib.parse import urlparse, urlunparse
+from six.moves.urllib.parse import urlunparse
 import argparse
 
 from posttroll.message import Message
 from posttroll.publisher import Publisher, get_own_ip
 from trollmoves.move_it_base import MoveItBase
-from trollmoves.client import Listener, request_push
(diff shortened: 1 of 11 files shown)


bin/move_it_mirror.py
```diff
@@ -20,59 +20,64 @@
...
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Move it Mirror."""
 
 import logging
 import logging.handlers
-import os
 from threading import Lock, Timer
-from six.moves.urllib.parse import urlparse, urlunparse
+from six.moves.urllib.parse import urlunparse
```
pnuu, line 28: We don't support Python 2 anymore, so `six` isn't needed.

trollmoves/mirror.py
```diff
@@ -0,0 +1,72 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""All you need for mirroring."""
+
+import os
+
+from six.moves.urllib.parse import urlparse
```
pnuu, line 27: We don't support Python 2 anymore, so `six` isn't needed.

trollmoves/mirror.py
```diff
@@ -0,0 +1,72 @@
...
+file_registry = {}
+
+
+class MirrorRequestManager(RequestManager):
+    """Handle requests as a mirror."""
+
+    def __init__(self, port, attrs):
+        """Set up this mirror request manager."""
+        RequestManager.__init__(self, port, attrs)
+        self._deleter = MirrorDeleter(attrs)
```
pnuu, line 42: The addition of `attrs` to this call is the actual fix and was pretty well hidden among other changes, so I'll just highlight it here.

trollmoves/tests/test_server.py
```diff
@@ -123,3 +124,12 @@ def test_process_notify(Message):
         with file_cache_lock:
             assert "/topic/20200428_1000_foo.tif" in file_cache
             assert len(file_cache) == 1
+
+
+class TestDeleter(unittest.TestCase):
```
pnuu, line 129: Should there be a test that the file is actually deleted?

pnuu (approved):
Only some comments, LGTM.

trollmoves/tests/test_server.py
mraspaud: There should indeed, but I didn't change anything about this in the PR...

trollmoves/mirror.py
mraspaud: thanks

Issue #85, opened 2020-12-17
Trollmoves Client shutsdown when config is modified


pnuu:
This happened when I removed one source from the `providers` line in `trollmoves_client_polar_ears_viirs.ini`:

```
[2020-12-17 14:21:32,552 DEBUG    trollmoves.client] New config file detected: /path/to/config_files/trollmoves_client_polar_ears_viirs.ini
[2020-12-17 14:21:32,554 DEBUG    posttroll.publisher] exiting publish
[2020-12-17 14:21:33,375 DEBUG    trollmoves.client] Exiting listener tcp://server1.fmi.fi:9011
[2020-12-17 14:21:33,762 DEBUG    trollmoves.client] Exiting listener tcp://server2.fmi.fi:9011
[2020-12-17 14:21:34,083 DEBUG    posttroll.publisher] Publish heartbeat (min_interval is 30.0 sec)
[2020-12-17 14:21:34,084 DEBUG    posttroll.publisher] exiting publish
[2020-12-17 14:21:34,600 DEBUG    trollmoves.client] Exiting listener tcp://sat7-io.fmi.fi:9011
[2020-12-17 14:21:35,557 DEBUG    trollmoves.client] Updating ears_suomi-npp_viirs
[2020-12-17 14:21:35,558 INFO     posttroll.publisher] publisher started on port 40000
[2020-12-17 14:21:35,559 DEBUG    posttroll.publisher] entering publish tcp://*:40000
[2020-12-17 14:21:35,560 DEBUG    trollmoves.client] Add listener for tcp://server1.fmi.fi:9011 with topic ['/ears/viirs/suomi-npp/1b', '/heartbeat/move_it_server']
[2020-12-17 14:21:35,561 DEBUG    trollmoves.heartbeat_monitor] Entering the Heartbeat monitor 0.00
[2020-12-17 14:21:35,561 DEBUG    trollmoves.client] Starting listener tcp://server1.fmi.fi:9011
[2020-12-17 14:21:35,561 INFO     trollmoves.client] Subscribing to tcp://server1.fmi.fi:9011 with topics ['/ears/viirs/suomi-npp/1b', '/heartbeat/move_it_server']
[2020-12-17 14:21:35,561 INFO     posttroll.subscriber] Subscriber adding address tcp://server1.fmi.fi:9011 with topics ['pytroll://ears/viirs/suomi-npp/1b', 'pytroll://heartbeat/move_it_server']
[...]

PR #95, opened 2021-10-22
Bugfix spare client requests

pnuu:
The current version of Trollmoves Client doesn't work properly when a hot-spare Client is used. The spare doesn't receive any of the necessary messages from the primary.

This PR:
- removes unnecessary publishers used for synchronization between the clients (these messages were never received)
- fixes the way different message types were handled
- restructures the unit tests for `trollmoves.client.Listener` class

There still is an issue with the Client stopping when the the config is modified and is reloaded. See https://github.com/pytroll/trollmoves/issues/85 for more information.

 - [x] Tests updated

### DIFF
--- a/bin/move_it_client.py
+++ b/bin/move_it_client.py
@@ -82,8 +82,6 @@ import argparse
 import signal
 import time
 
-from posttroll.publisher import NoisyPublisher
-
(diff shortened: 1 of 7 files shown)


pnuu:
Approved on Slack by @mraspaud, merging.

PR #103, opened 2021-11-29
Remove client listener callback

pnuu:
In a continued effort to make Trollmoves easier to maintain, this PR removes the callback mechanism from the Client. In the same go, the unused `trollmoves.client.StatCollector` is removed.

 - [x] Tests updated


mraspaud (approved):
LGTM!

PR #104, opened 2021-11-30
Refactor client

pnuu:
Yet another step in refactoring Trollmoves. This PR cleans the `trollmoves.client` code base to use smaller methods/functions.

 - [x] Tests updated

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -196,77 +196,67 @@ class Listener(Thread):
         """Run listener."""
         try:
             with heartbeat_monitor.Monitor(self.restart_event, **self.ckwargs) as beat_monitor:
-
                 self.running = True
-
                 while self.running:
-                    # Loop for restart.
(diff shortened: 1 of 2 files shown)


trollmoves/client.py
```diff
@@ -254,6 +265,36 @@ def stop(self):
             self.subscriber = None
 
 
+def _is_push_message(msg):
```
mraspaud, line 268: These functions are not only checking, but also acting on the message. Hence I think the `is_...` formulation is misleading. Maybe `_process_push_message` would be more appropriate?

pnuu: Yeah, I was thinking the same, but didn't come up with a better name. I think the `_process` prefix is better in any case, so I'll use that.

pnuu: I went with `_handle_*`, then the `is_message_already_handled()` method matches the naming.

PR #105, opened 2021-12-01
Refactor server

pnuu:
Another refactoring PR, now for the Server code.

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -255,20 +255,9 @@ class RequestManager(Thread):
 
     def info(self, message):
         """Collect information from file cache to message."""
-        topic = message.subject
-        max_count = 2256  # Let's set a (close to arbitrary) limit on messages size.
-        try:
-            max_count = min(message.data.get("max_count", max_count), max_count)
-        except AttributeError:
-            pass
         uptime = datetime.datetime.utcnow() - START_TIME
-        files = []
-        with file_cache_lock:
-            for i in file_cache:
-                if i.startswith(topic):
-                    files.append(i)
-                    if len(files) == max_count:
-                        break
+        files, max_count = _collect_cached_files(message)
+
         return Message(message.subject, "info", data={"files": files, "max_count": max_count, "uptime": str(uptime)})
 
     def unknown(self, message):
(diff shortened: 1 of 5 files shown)


trollmoves/server.py
```diff
@@ -418,32 +463,29 @@ def process_notify(orig_pathname, publisher, pattern, attrs):
...
-    elif (os.stat(orig_pathname).st_size == 0):
-        # Want to avoid files with size 0.
+    elif os.stat(orig_pathname).st_size == 0:
         LOGGER.debug("Ignoring empty file: %s", orig_pathname)
         return
     else:
         LOGGER.debug('We have a match: %s', orig_pathname)
 
     pathname = unpack(orig_pathname, **attrs)
+    info = _get_inotify_message_info(attrs, orig_pathname, pathname)
```
mraspaud, line 473: I'm a bit confused by the function name here, but maybe I'm not really following: what does inotify have to do in the posttroll notifier function?

mraspaud (commented):
Thanks a lot for the huge refactoring work! Just one clarification needed for my part

trollmoves/server.py
pnuu: Umm, the `_get_inotify_message_info()` is called in `process_notify()` function which is called only in the nested function `process_notify_publish()` of `create_inotify_notifier()`. So it is part of the inotify chain :-) I'll adjust the ordering of the functions to make this more clear.

pnuu: Oh, you were right. It is used as a callback (sigh) in all the `create_*_notifier()` functions.

trollmoves/server.py
```diff
@@ -192,7 +213,7 @@ def push(self, message):
         new_msg = self._move_files(message)
         if new_msg is None:
             new_msg = Message(message.subject,
-                              message.type,
+                              'file',
```
mraspaud, line 216: Don't we have `dataset` types of messages coming here?

mraspaud (commented):
One question, otherwise LGTM

trollmoves/server.py
pnuu: Could be? I don't remember seeing those in my own setups. If you have such a message, I could perhaps adapt this. Or would it be sufficient to check the message data for `'uri'`, `'dataset'` and `'collection'` and decide based on that? The incomming message in any case has type `'push'`.

pnuu: Oh, I guess @TAlonglong had one example in an issue, I'll have a look after lunch.

pnuu: Ok, done.

PR #106, opened 2021-12-01
Refactor mirror

pnuu:
Refactoring effort to make Trollmoves Mirror to work with already refactored Trollmoves Client.

Also adds the `MoveItMirror.run()` method that has been absent since version 0.6.0. So last usable version with Mirror at the moment is 0.6.0.

### DIFF
--- a/bin/move_it_mirror.py
+++ b/bin/move_it_mirror.py
@@ -23,120 +23,12 @@
 """Move it Mirror."""
 
 import logging.handlers
-from threading import Lock, Timer
-from urllib.parse import urlunparse
 import argparse
 
-from posttroll.message import Message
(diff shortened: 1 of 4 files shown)


pnuu:
I need to setup a Server - Mirror - Client chain locally to be certain *anything* works in the Mirror. There were practically no tests, and some of the functional parts were outside of the library, so I didn't notice the `callback` structure of Trollmoves Client were used here when refactoring Client.

mraspaud (approved):
LGTM

pnuu:
I hope this will also fix the bug shown in #107.

pnuu:
And I moved this back to "Ready for review", as the previous review was for incomplete work.

PR #108, opened 2021-12-02
Refactor dispatcher

pnuu:
The last module-level refactoring. This also includes a fix for the Client not publishing a meaningful message due to a missing `return`.


pnuu:
Converted to draft, there are still bits to refactor.

pnuu:
Bah, already forgot that I'd changed to `main` and it looked like there was a lot to do. This is now ready for review.

PR #109, opened 2021-12-02
Example end-to-end test

pnuu:
This PR adds an example how the chain from `Trollmoves Server` to `Trollmoves Client` via a `Trollmoves Mirror` works, and further from `Client` to `Dispatcher`, works. This can be used to test that everything works, and as a starting point to set up file transfers. Usage will require #105, #106 and #108 to have been merged.


mraspaud (approved):
Very nice file contents, LGTM :)

PR #113, opened 2022-03-02
Fix dispatcher destination url creation and revert test config to intended value

mraspaud:
Some functionality was accidently removed during the last refactoring, not allowing eg a directory to be provided in the dispatcher config for each config item but not providing any default. This PR fixes it, and uses the correct yaml test config for testing this.

 - [x] Tests fixed

### DIFF
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -388,19 +388,18 @@ class Dispatcher(Thread):
 
     def create_dest_url(self, msg, client, conf):
         """Create the destination URL and the connection parameters."""
-        defaults = self.config[client].copy()
-        _verify_filepattern(defaults, msg)
-        conf_with_defaults = defaults.copy()
-        conf_with_defaults.update(conf)
-        connection_parameters = conf_with_defaults.get('connection_parameters')
+        config = self.config[client].copy()
+        _verify_filepattern(config, msg)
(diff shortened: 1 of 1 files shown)


trollmoves/dispatcher.py
```diff
@@ -390,14 +390,17 @@ def create_dest_url(self, msg, client, conf):
         """Create the destination URL and the connection parameters."""
         defaults = self.config[client].copy()
         _verify_filepattern(defaults, msg)
-        connection_parameters = conf.get('connection_parameters', defaults.get('connection_parameters'))
-        host = conf.get('host', defaults['host'])
+        conf_with_defaults = defaults.copy()
```
pnuu, line 393: I think this copy is unnecessary. Yes, it makes the naming more clear, but still unnecessary copy since `defaults` is not used after this.

pnuu (approved):
LGTM, just one comment.

trollmoves/dispatcher.py
mraspaud: Ok, would you be ok with this?
```suggestion
        conf_with_defaults = defaults
```

pnuu: Yes, or rename `defaults` to something more generic as `configuration` so the extra variable isn't needed.

PR #114, opened 2022-04-11
Refactoring move it server/client and allow passing fsspec-base remote info

mraspaud:
This PR refactors the move it client/mirror/server to be more OOP complient, allowing easier future extensions.

This also adds some tests for the move it modules, in particular some acceptance tests doing some actual file-moving.

An extension is thus added here also to allow publishing messaged from the move it server side that include fsspec complient remote-filesystem specifications, for later use in trollflow2 and satpy.

This needs https://github.com/pytroll/pytroll-collectors/pull/104

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pytest-bdd pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors
+          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors
 
       - name: Install trollmoves
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog
+          pip install -U codecov pytest pytest-cov pytest-bdd pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog
 
       - name: Install trollmoves
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors fsspec
+          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors>=0.13.0 fsspec
 
       - name: Install trollmoves
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pytest-bdd pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog
+          pip install -U codecov pytest pytest-cov pytest-bdd pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors
 
       - name: Install trollmoves
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors
+          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors fsspec
 
       - name: Install trollmoves
         run: |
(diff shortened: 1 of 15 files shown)


mraspaud:
This depends on https://github.com/pytroll/pytroll-collectors/pull/104

trollmoves/server.py
```diff
@@ -692,21 +739,40 @@ def process_notify(orig_pathname, publisher, pattern, attrs):
         LOGGER.debug('We have a match: %s', orig_pathname)
 
     pathname = unpack(orig_pathname, **attrs)
+    if "request_port" in attrs:
+        msg = create_message_with_request_info(pathname, orig_pathname, attrs)
+    else:
+        msg = create_message_with_remote_fs_info(pathname, attrs)
```
pnuu, line 745: So if the `request_port` isn't defined, it means that there won't be a need for a matching Client to make transfer requests? And then the consumer (Trollflow2/Satpy) would read then listen to directly the messages sent by the Server process, and need to read the data directly from the original source (reception server or such) using `fsspec` and have the required credentials available at their end?

trollmoves/server.py
```diff
@@ -603,53 +628,75 @@ def _chains_are_identical(chains, new_chains, chain_name):
...
+#         chain["notifier"].join()
+#     except AttributeError:
+#         pass
+#     if "request_manager" in chain:
+#         chain["request_manager"].stop()
+#         LOGGER.debug('Stopped the request manager')
+
+
+class Chain:
+    """A chain."""
```
pnuu, line 650:
```suggestion
    """A chain for managing new incoming files."""
```

pnuu (commented):
Thanks for the refactoring!

At this stage only few questions and one suggestion for a clearer docstring. Didn't want to try understanding what these new tests are supposed to be doing, but maybe some day.

.github/workflows/ci.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U codecov pytest pytest-cov pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog
+          pip install -U codecov pytest pytest-cov pytest-bdd pytest-reraise pyyaml trollsift posttroll inotify pyinotify paramiko scp watchdog pytroll-collectors
```
pnuu, line 30: `pytroll-collectors` needs to be added as an optional dependency in `setup.py`, and `pytest-bdd` and `pytest-reraise` as test requirements.

trollmoves/server.py
mraspaud: yes, this is correct.

.github/workflows/ci.yaml
mraspaud: thanks, adding

mraspaud: hmm, the documentation of setuptools says `tests_require` is deprecated (since `python setup.py test` is deprecated)... I'm not sure how to add the tests requirements now.

pnuu:
New pytroll-collectors with https://github.com/pytroll/pytroll-collectors/pull/104 included has been released so the CI should now work. Well, after the conflicts have been checked at least.

pnuu:
Some tests for `MoveItMirror` are failing.

mraspaud:
@pnuu tests passing now. Any more comments?

trollmoves/tests/test_client.py
```diff
@@ -976,16 +959,15 @@ def test_read_config(client_config_1_item):
...
+                "nameservers", "providers", "topic", "publish_port", ]:
         assert key in section_keys
     assert isinstance(conf[section_name]["providers"], list)
 
 
 @patch('trollmoves.client.request_push')
 @patch('trollmoves.client.create_publisher_from_dict_config')
 @patch('trollmoves.client.Listener')
-def test_reload_config_single_chain(Listener, create_publisher_from_dict_config, request_push, client_config_1_item):
+def test_reload_config_single_chain(Listener, NoisyPublisher, request_push, client_config_1_item):
```
pnuu, line 970: Why are these renamed to `NoisyPublisher`?

pnuu (commented):
Thanks for the refactoring. Few questions and suggestions inline. There were so many changes that seeing the full new structure from the GitHub diffs is pretty much impossible, but I guess it's better now.

Also, there should be some documentation on the new fsspec stuff. Maybe add them to #130 ?

mraspaud:
@pnuu regarding 130, should we merge it so that we can add bits in other PRs around it then?

trollmoves/tests/test_client.py
mraspaud: that's a miss, fixing

pnuu:
Yes, I think that'd be good.

trollmoves/tests/test_client.py
mraspaud: done

trollmoves/move_it_base.py
```diff
@@ -104,35 +86,52 @@ def setup_watchers(self, cmd_args):
...
-        logger_name = "move_it_client"
-    elif chain_type == "mirror":
-        logger_name = "move_it_mirror"
-    LOGGER = logging.getLogger(logger_name)
-    pyinotify.log.handlers = [fh_]
+        self.watchman.add_watch(os.path.dirname(self.cmd_args.config_file), mask)
+
+    def setup_logging(self):
+        """Set up logging."""
+        global LOGGER
```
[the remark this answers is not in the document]
mraspaud: Ok, this was just moved and was already like this in the former place.
Actually, I'm thinking that the logger should be independent of the move it classes, so I just started refactoring this bit to have a unified logging utility for all of trollmoves. But I realise this is probably best to be provided in another PR after we merge this one, what do you think @pnuu ?

mraspaud:
Documentation added. I think I addressed all your concerns @pnuu ?

trollmoves/move_it_base.py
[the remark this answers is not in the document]
pnuu: Agreed, better to do in a separate PR.

pnuu (approved):
LGTM!

PR #115, opened 2022-04-26
Add a mover for s3:// protocol

pnuu:
This PR adds a mover for `s3://` protocol, that is to S3 object storage. Some related adjustments were done to Trollmoves Client, including previously missing unit tests.

 - [x] Tests added  
 - [x] Fully documented

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -673,7 +673,6 @@ def _send_ack_message(msg, publisher):
 
 
 def _update_local_message(lmsg, _destination, login, response, **kwargs):
-    scheme = urlparse(_destination).scheme
     lmsg = make_uris(lmsg, _destination, login)
     lmsg.data['origin'] = response.data['request_address']
     lmsg.data.pop('request_address', None)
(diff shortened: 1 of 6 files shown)


trollmoves/client.py
```diff
@@ -470,7 +472,7 @@ def make_uris(msg, destination, login=None):
     duri = urlparse(destination)
     scheme = duri.scheme or 'ssh'
     dest_hostname = duri.hostname or socket.gethostname()
-    if socket.gethostbyname(dest_hostname) in get_local_ips():
+    if scheme not in ('s3') and socket.gethostbyname(dest_hostname) in get_local_ips():
```
mraspaud, line 475: I'm not really found of listing "s3" here, as we would need to add also "gcs" if we add it as a mover in the future. But maybe this is premature optimization, to do what you want with this comment.

mraspaud (approved):
LGTM

trollmoves/client.py
pnuu: I kinda expected there would be similar cases so did `not in ('s3')` instead of `!= 's3'` :grin: .

PR #119, opened 2022-05-17
Add an option to disable target directory creation

pnuu:
I'd want to move the Clients to servers that don't have access to the target filesystem. This isn't possible at the moment because target path creation is always attempted. This PR adds a new configuration option that can be used to disable it.

 - [x] Tests added  
 - [ ] Fully documented

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -101,15 +101,22 @@ def _set_config_defaults(conf):
     conf.setdefault("req_timeout", DEFAULT_REQ_TIMEOUT)
     conf.setdefault("transfer_req_timeout", 10 * DEFAULT_REQ_TIMEOUT)
     conf.setdefault("nameservers", None)
+    conf.setdefault("create_target_directory", True)
+
+
+FALSY = ["", "False", "false", "0", "off"]
+TRUTHY = ["True", "true", "on", "1"]
 
 
 def _parse_boolean_config_items(conf):
-    if conf["delete"] in ["", "False", "false", "0", "off"]:
+    if conf["delete"] in FALSY:
         conf["delete"] = False
-    if conf["delete"] in ["True", "true", "on", "1"]:
+    if conf["delete"] in TRUTHY:
         conf["delete"] = True
-    if conf["heartbeat"] in ["", "False", "false", "0", "off"]:
+    if conf["heartbeat"] in FALSY:
         conf["heartbeat"] = False
+    if conf["create_target_directory"] in FALSY:
+        conf["create_target_directory"] = False
 
 
 def _check_provider_config(conf, section):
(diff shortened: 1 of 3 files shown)


pnuu:
`pre-commit` wasn't happy with the fixture definitions so I used `--no-verify` in the commit.

mraspaud:
I'm not really convince this is the best solution. The way I understand the problem, the destination directory will be created at some point by someone else. So how about just checking if the directory exists before creating it?

pnuu:
The client _can't_ do the check because it doesn't exist on the host(s) where the client(s) is/are located:
- two reception servers, each having a Trollmoves Server for any given data
- the same servers have a Trollmoves Client each (one of them configured as hot spare)
- third, completely separate server where the data are copied to, the filesystem not directly accessible by the Client (not mounted at the server they are running on)

mraspaud:
Sorry, I don't understand. How does the Client put the data where it is configured to, when it doesn't have access to the filesystem at all?

pnuu:
It can do the transfer via `scp`, but can't check the existence of the directory using `os.path.exists()` as the filesystem is not mounted on the server.

mraspaud:
oooh, ok, sorry, I understand now. Yeah, it's a bit convoluted. Anyway, I still think having an option for this is overkill, a simple try except around the directory creation with a debug log message in case of failure is enough. What do you think?

pnuu:
That's what's there in the current version, although in addition there would be a need to do additional network name and/or IP address checks, and that sounds too complicated. The config option is the only way I could think of that was simple to implement.

mraspaud:
I can't see try/except around the `create_local_directory`, sorry, can you point me to the try/except? Quite tired those days...

pnuu:
But yeah, the try/except would also need the host check, and that's more complicated and less explicit.

mraspaud:
OK

trollmoves/client.py
```diff
@@ -615,7 +624,8 @@ def _request_files(huid, destination, login, publisher, **kwargs):
 
         req, fake_req = create_push_req_message(msg, _destination, login)
         LOGGER.info("Requesting: %s", str(fake_req))
-        local_dir = create_local_dir(_destination, kwargs.get('ftp_root', '/'))
+        local_dir = create_local_dir(_destination, kwargs.get('ftp_root', '/'),
+                                     create_target_directory=kwargs.get('create_target_directory', True))
```
mraspaud, line 628: I'd rather have the test here, otherwise it will be a bit strange, right?

```python
...
result = do_something(args, do_it=False)
...
```

mraspaud (changes requested):
Just one request, otherwise it looks good.

trollmoves/client.py
pnuu: Done.

mraspaud (approved):
LGTM

PR #121, opened 2022-06-17
added configurable scp put timeout. added tests

### LINKED ISSUE #120: Add configurable timeout to SCPClient
Default timeout to the SCPClient channel when using put is 10 seconds. The default is hardcoded in the SCPClient code https://github.com/jbardin/scp.py/blob/8d8ffd018f7f2f34dc99b413cf3f817705cc9233/scp.py#L117

This should be possible to configure in the move_it_server configure section and pass into the SCPClient

I suggest to add a `scpclient_timeout_seconds` as an option in the config file.

So add this in the SCPMover and handle accordingly.

TAlonglong:
Added configurable scpclient_timeout_seconds attribute to the scpclient put.

 - [X] Closes #120  
 - [X] Tests added  
 - [ ] Fully documented


TAlonglong:
Not sure what the stickler issue is about. ELLIPSIS_MARKER is not used. Was it used in the past and not used anymore?

If you have time @pnuu

trollmoves/movers.py
```diff
@@ -23,6 +23,7 @@
 
 """Movers for the move_it scripts."""
 
+from doctest import ELLIPSIS_MARKER
```
pnuu, line 26: I don't see this in `main` at all, so don't know where it's coming from. Should be safe to remove.

trollmoves/tests/test_ssh_server.py
```diff
@@ -178,6 +179,18 @@ def test_scp_copy(self, mock_scp_client, mock_sshclient):
 
         mocked_scp_client.put.assert_called_once_with(self.origin, urlparse(self.destination_no_port).path)
 
+    @patch('trollmoves.movers.SSHClient', autospec=True)
+    @patch('trollmoves.movers.SCPClient', autospec=True)
+    def test_scp_copy_custom_timeout(self, mock_scp_client, mock_sshclient):
```
pnuu, line 184: The timeout isn't tested in this test, so either adjust the name or test the timeout.

pnuu (changes requested):
In principle LGTM, but couple of requests inline.

trollmoves/movers.py
pnuu: Oh, actually, it's coming from this PR since it shows up in the diff.

pnuu:
I asked Claude to bring this PR up-to-date with current `main`. Here are the it did:

Main had refactored ScpMover substantially since the branch's merge base: scp/paramiko imports moved into the methods that use them, connection and transfer failures are now retried through _run_with_retries(), and the broad except Exception handlers were narrowed. The configurable timeout was reshaped onto that structure rather than kept as written:

  - movers.py:560 — a _scpclient_timeout() accessor sits next to the existing _ssh_connection_timeout() / _num_ssh_retries(), so an ini-config value arriving as a string gets converted like its siblings do.
  - movers.py:650 — the scpclient_timeout_seconds hint is logged from main's combined (SCPException, SSHException) handler instead of a separate SCPException handler. Keeping the branch's handler would have shadowed the retry path, so a timeout would have stopped being retried.
  - Two bugs from the branch are gone: str(scpe) in "Timeout waiting for scp response" was a substring test the wrong way round (it matched only short fragments, never the real message), and from doctest import ELLIPSIS_MARKER was an accidental auto-import.
 
Tests:
Rewritten against main's patch targets (scp.SCPClient / paramiko.SSHClient, now that the imports are lazy). The original timeout test only asserted put was called, so it passed with or without the feature — the replacements assert the timeout actually reaches SCPClient, including the string-valued case, plus one that a timed-out transfer logs the hint. The branch's two SCPException tests were dropped: main already covers SCPException propagation and retrying.

pnuu:
Claude identified that some of the raised timeouts were caught in the wrong place and gave incorrect log messages. This is fixed in 3b7c61c

pnuu (approved):
LGTM

PR #126, opened 2022-08-31
Make libraries required only by Trollmoves Server optional extras

pnuu:
To reduce the size of a Trollmoves Client container image, make the libraries used only by the Server optional extras.

This also means that a full installation would need `pip install 'trollmoves[all]'` or omitting S3 parts `pip install 'trollmoves[server]'`.

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -35,16 +35,6 @@ import netrc
 
 from ftplib import FTP, all_errors, error_perm
 try:
-    from paramiko import SSHClient, SSHException, AutoAddPolicy
-except ImportError:
-    SSHClient = None
-    SSHException = None
-    AutoAddPolicy = None
-try:
-    from scp import SCPClient
-except ImportError:
-    SCPClient = None
-try:
     from s3fs import S3FileSystem
 except ImportError:
     S3FileSystem = None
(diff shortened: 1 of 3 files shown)


trollmoves/movers.py
```diff
@@ -297,6 +306,8 @@ class ScpMover(Mover):
 
     def open_connection(self):
         """Open a connection."""
+        if SSHClient is None or SCPClient is None:
```
mraspaud, line 309: why not import SSH Client here directly and let it raise the error if not found?

mraspaud: (and same for SCPClient where it's used)

mraspaud (approved):
LGTM, just some minor suggestion

trollmoves/movers.py
pnuu: I thought it better to put these like this so that also the `SCPClient` import was tested already when opening the connection. It is accessed only when `copy()` is called, so the failure can be delayed for a significant time (e.g. with polar satellite data) after the startup. *BUT* now that I read the code, the call to `copy()` actually has the connection creation as the first step, so this will happen in any case.

I'll move the imports as suggested.

pnuu: Done. Needed to add some mocking in the tests.

PR #127, opened 2022-08-31
Use dictionary config interfaces for publisher and subscriber creation

pnuu:
This PR makes it possible to:
- disable `nameserver` connections by defining `nameservers = False` (Client) or `nameserver = False` and `addresses = srvr1:12345 srvr2:12345` (Server, when using posttroll notifier) in the configuration file
- pass all the `NSSubscriber` parameters by defining them in the Server configuration file when using posttroll notifier

The default behavior isn't affected.

 - [x] Tests adjusted  
 - [ ] Fully documented

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -41,7 +41,7 @@ from zmq import LINGER, POLLIN, REQ, Poller
 import bz2
 from posttroll import get_context
 from posttroll.message import Message, MessageError
-from posttroll.publisher import NoisyPublisher
+from posttroll.publisher import create_publisher_from_dict_config
 from posttroll.subscriber import Subscriber
 from trollsift.parser import compose
 
@@ -700,7 +700,6 @@ class Chain(Thread):
         super(Chain, self).__init__()
         self._config = config
         self._name = name
-        self._np = None
         self.publisher = None
         self.listeners = {}
         self.listener_died_event = Event()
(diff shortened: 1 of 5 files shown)


trollmoves/server.py
```diff
@@ -428,7 +428,22 @@ def __init__(self, attrs, publisher):
 
     def run(self):
         """Start listening to messages."""
-        with Subscribe('', topics=self.attrs['listen'], addr_listener=True) as sub:
+        nameserver = self.attrs.get('nameserver')
+        if nameserver in ('false', 'False'):
```
mraspaud, line 432: why not use `FALSY` here?

trollmoves/server.py
```diff
@@ -428,7 +428,22 @@ def __init__(self, attrs, publisher):
...
+
+        with Subscribe(
+            services=self.attrs.get('services', ''),
+            topics=self.attrs.get('topics', self.attrs['listen']),
+            addr_listener=bool(self.attrs.get('addr_listener', True)),
+            addresses=addresses,
+            timeout=int(self.attrs.get('timeout')),
+            translate=bool(self.attrs.get('translate', False)),
+            nameserver=nameserver,
+        ) as sub:
```
mraspaud, line 446: Is there a test to check at least that the old behaviour is unchanged given the same config?

mraspaud: Asking because codecov says that parts of this function are untested...

trollmoves/client.py
```diff
@@ -709,16 +708,19 @@ def __init__(self, name, config):
 
     def setup_publisher(self):
         """Initialize publisher."""
-        if self._np is None:
+        if self.publisher is None:
             try:
                 nameservers = self._config["nameservers"]
+                if nameservers in FALSY:
+                    nameservers = False
```
mraspaud, line 715: Codecov says this is not tested?

mraspaud (commented):
The logic LGTM, some testing questions

trollmoves/server.py
pnuu: I'll move `FALSY` to `__init__` and use it from there where needed.

trollmoves/server.py
pnuu: Seems that the posttroll notifier (`trollmoves.server.Listener` class) has no tests. I'll see if I can figure out enough of it to create a test to see the default usage hasn't changed.

trollmoves/server.py
pnuu: Done

trollmoves/server.py
pnuu: I added the bare minimum test for this.

pnuu:
I refactored the code and config reading a bit so that adding YAML config support would be a bit easier.

mraspaud (commented):
Looks good, just a couple of small test to exercise the two `read_config` functions and I'm happy :smile:

mraspaud (approved):
LGTM, thanks for taking the time to refactor the config parsing.

PR #129, opened 2022-09-09
Separate install requirements for 3rd party packages

pnuu:
This PR is an attempt to make it possible to use as small portions of Trollmoves as possible.

As an example, https://github.com/pytroll/pytroll-collectors/pull/117 would only require the `S3Mover` with its install requirements, so installation of all the other 3rd party libraries would be unnecessary.

Another PR that starts the documentation: https://github.com/pytroll/trollmoves/pull/130

 - [ ] Fully documented

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ extras_require = {
     'dispatcher': [
         'inotify',
         'netifaces',
+        'posttroll>=1.5.1',
         'pyyaml',
         'trollsift',
     ],
--- a/setup.py
+++ b/setup.py
@@ -26,14 +26,44 @@ import versioneer
 
 
 extras_require = {
+    'client': [
+        'netifaces',
+        'posttroll>=1.5.1',
+        'pyinotify',
+        'pyzmq',
+        'trollsift',
+    ],
+    'dispatcher': [
+        'inotify',
+        'netifaces',
+        'pyyaml',
+        'trollsift',
+    ],
+    'mirror': [
+        'netifaces',
+        'posttroll>=1.5.1',
+        'pyinotify',
+        'pyzmq',
+        'trollsift',
+        'watchdog',
+    ],
     's3': [
         's3fs',
     ],
-    'server': [
-        'inotify',
+    'scp': [
         'paramiko',
         'scp',
+    ],
+    'server': [
+        'netifaces',
+        'posttroll>=1.5.1',
+        'pyinotify',
+        'pyzmq',
+        'trollsift',
         'watchdog',
+    ],
+    'sftp': [
+        'paramiko',
     ],
 }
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -26,15 +26,45 @@
...
 
 extras_require = {
+    'client': [
+        'netifaces',
+        'posttroll>=1.5.1',
+        'pyinotify',
+        'pyzmq',
+        'trollsift',
+    ],
+    'dispatcher': [
```
mraspaud, line 36: dispatcher needs posttroll too to receive messages

trollmoves/utils.py
```diff
@@ -19,10 +19,16 @@
...
+"""Utility functions for Trollmoves."""
+
 import socket
 from urllib.parse import urlparse, urlunparse
 
-import netifaces
+try:
+    import netifaces
+except ImportError:
+    netifaces = None
```
mraspaud, line 31: I really don't like this, sorry. Could we just move `get_local_ips` to client.py instead, where we know netifaces are a requirement?

setup.py
```diff
@@ -26,15 +26,45 @@
 
 
 extras_require = {
+    'client': [
+        'netifaces',
+        'posttroll>=1.5.1',
+        'pyinotify',
+        'pyzmq',
+        'trollsift',
+    ],
```
mraspaud, line 35: Looks like ther requirements for client, mirror, server, dispatcher are pretty similar, do we really need to separate them?

mraspaud: also, they seem to be the core of the package, why make them extra requirements? I can understand making optional for movers, since only the user know what's going to be used (s3, scp, paramiko)

mraspaud (changes requested):
Thanks for separating the movers requirements. However, I think we need to talk about what is core requirements for the entire package.

setup.py
pnuu: If there's another way for this I'm all ears. Client and dispatcher have the same requirements, as do mirror and server which require watchdog in addition to client and dispatcher.

setup.py
pnuu: Fixed in https://github.com/pytroll/trollmoves/pull/129/commits/f1c5fbfc41b2935e8a0be701a4b608b1cf2eadf4

trollmoves/utils.py
pnuu: Fixed in https://github.com/pytroll/trollmoves/pull/129/commits/096231c40ac185207d7fa3827bc9b881ce6c41a5

setup.py
mraspaud: can't we have this simply as `install_requires`, including watchdog? Then we cover the base needs for 90% of the code...

pnuu: But we can't then use the S3 mover without installing all the unnecessary requirements. For cloud and microservices that's a no-go.

mraspaud: aah, ok, so I suppose that's a sign that trollmoves and movers should be two separate packages...

pnuu: Yes, that might be one solution :thinking:

pnuu:
Duh. Installing the dependencies required for a specific use-case and doing `pip install --no-dependencies trollmoves` achieves the same without making the full installation harder (`pip install trollmoves[all]`).

Closing.

PR #134, opened 2022-11-02
Import `netifaces` only if used

pnuu:
This PR moves the `netifaces` import inside the `trollmoves.utils.get_local_ips()` function. This way `pip install --no-dependencies trollmoves` can be used with minimal additional packages when `get_local_ips()` isn't needed.

There were also some docstrings missing, which are added here.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 


mraspaud (approved):
LGTM!

PR #140, opened 2022-11-21
Add try/except around Message creation

TAlonglong:
Add try/except around Message creation in request manager loop.

This to avoid the request manager loop to crash if it get an invalid message.

 - [x] Tests added  
 - [x] Fully documented 

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -233,23 +233,26 @@ class RequestManager(Thread):
     def _run(self):
         """Run request manager."""
         while self._loop:
+            self._run_loop
+
+    def _run_loop(self):
+        """Run request manager loop."""
+        try:
+            socks = dict(self._poller.poll(timeout=2000))
+        except ZMQError:
+            LOGGER.info("Poller interrupted.")
+            return
+        if socks.get(self.out_socket) == POLLIN:
+            address, payload = self._get_address_and_payload()
+            if payload is None:
+                return
             try:
-                socks = dict(self._poller.poll(timeout=2000))
-            except ZMQError:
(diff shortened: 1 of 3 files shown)


trollmoves/server.py
```diff
@@ -242,7 +242,12 @@ def _run(self):
                 address, payload = self._get_address_and_payload()
                 if payload is None:
                     continue
-                self._process_request(Message(rawstr=payload), address)
+                try:
+                    self._process_request(Message(rawstr=payload), address)
+                except MessageError:
+                    LOGGER.exception("Failed to create message from payload: %s with address %s",
+                                     str(payload), str(address))
+                    pass
```
mraspaud, line 250:
```suggestion
```

mraspaud (commented):
Could we add a test for this?

TAlonglong:
Working on the test now.

Why not pass? I know its not needed, but it clearly indicates that the intention is not the break the loop

mraspaud:
> Working on the test now.
> 
> Why not pass? I know its not needed, but it clearly indicates that the intention is not the break the loop

`pass` is usually just written to avoid a syntax error, like in
```python
def dummy_function():
    pass
```
It's unusual in python to use it as you do, so I thought it was something that you forgot to clean up. I would still remove it though, as the flow is clear to me without it.

TAlonglong:
Hm, tests passes locally, but not in CI.

Looks like caplog is empty whereas I expected some output. hm

TAlonglong:
yay,

mraspaud (approved):
LGTM!

PR #147, opened 2022-12-12
Add timeout to ssh client connect

### LINKED ISSUE #146: Add timeout to ssh connect
I have seen ssh connect get stuck when, for some unknown reason, the destination server can not be reached.

This cause the connect to block for a undefined length of time, pending on the underlying system I think. I have seen like 4 minutes before it returns with an exception. And it try this 3 times.

When this happens none of the other clients get their request handled, and when the connection finally get to the next client the file is deleted by the server and any transfer fails.

I want to add a timeout in the SSHClient connect.

Debugging this I to like this:
```
>>> from trollmoves import movers
>>> m = movers.ScpMover("/tmp/test", "scp://ubuntu@<test-address>:22/home/ubuntu")
>>> m.open_connection()
<paramiko.client.SSHClient object at 0x7f1522682850>
>>> m = movers.ScpMover("/tmp/test", "scp://ubuntu@<test-address>:23/home/ubuntu")
>>> m.open_connection()
Unknown exception at init SSHClient: [Errno 110] Connection timed out
Traceback (most recent call last):
  File "/home/trygveas/pytroll/trollmoves/trollmoves/movers.py", line 310, in open_connection
    ssh_connection.connect(self.destination.hostname,
  File "/home/trygveas/miniconda3/envs/trollmoves/lib/python3.9/site-packages/paramiko/client.py", line 358, in connect
    retry_on_signal(lambda: sock.connect(addr))
  File "/home/trygveas/miniconda3/envs/trollmoves/lib/python3.9/site-packages/paramiko/util.py", line 279, in retry_on_signal
    return function()
  File "/home/trygveas/miniconda3/envs/trollmoves/lib/python3.9/site-packages/paramiko/client.py", line 358, in <lambda>
    retry_on_signal(lambda: sock.connect(addr))
TimeoutError: [Errno 110] Connection timed out

times 3
```

I'm not 100% this port trick simulates what actually happened, but I looks very like so I will give it a try.

TAlonglong:
- [X] Closes #146 x 
 - [x] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -32,7 +32,6 @@ import time
 import traceback
 import socket
 from ftplib import FTP, all_errors, error_perm
-from paramiko import SSHClient, SSHException
 from threading import Event, Lock, Thread, current_thread
 from urllib.parse import urlparse
 
@@ -298,10 +297,11 @@ class ScpMover(Mover):
 
     def open_connection(self):
         """Open a connection."""
+        from paramiko import SSHClient, SSHException
 
         retries = 3
         ssh_key_filename = self.attrs.get("ssh_key_filename", None)
-        timeout = self.attrs.get("timeout", None)
+        timeout = self.attrs.get("ssh_connection_timeout", None)
         while retries > 0:
             retries -= 1
             try:
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -30,6 +30,7 @@ import shutil
 import sys
 import time
 import traceback
+import socket
 from ftplib import FTP, all_errors, error_perm
 from threading import Event, Lock, Thread, current_thread
 from urllib.parse import urlparse
@@ -300,6 +301,7 @@ class ScpMover(Mover):
 
         retries = 3
         ssh_key_filename = self.attrs.get("ssh_key_filename", None)
+        timeout = self.attrs.get("timeout", None)
         while retries > 0:
             retries -= 1
             try:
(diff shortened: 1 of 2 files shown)


trollmoves/movers.py
```diff
@@ -296,10 +298,10 @@ class ScpMover(Mover):
 
     def open_connection(self):
         """Open a connection."""
-        from paramiko import SSHClient, SSHException
 
         retries = 3
         ssh_key_filename = self.attrs.get("ssh_key_filename", None)
+        timeout = self.attrs.get("timeout", None)
```
mraspaud, line 304: I think the config variable name should be "ssh_connection_timeout", since it only applies to the ssh connection, right?

trollmoves/movers.py
```diff
@@ -30,7 +30,9 @@
 import sys
 import time
 import traceback
+import socket
 from ftplib import FTP, all_errors, error_perm
+from paramiko import SSHClient, SSHException
```
mraspaud, line 35: So paramiko is not required in setup.py at the moment, which will cause the import of movers.py fail in a fresh install of trollmoves.
One possibility would be to add paramiko the requirements (as it is now only an extra).
However, if possible I would avoid it, because paramiko in not systematically needed (eg if using FTP).
And also @pnuu has been trying to reduce the number of dependencies...

trollmoves/tests/test_ssh_server.py
```diff
@@ -149,7 +153,30 @@ def test_scp_open_connection_ssh_exception(self, mock_sshclient_connect):
...
             scp_mover.open_connection()
 
-    @patch('paramiko.SSHClient.connect', autospec=True)
+    @patch('trollmoves.movers.SSHClient.connect', autospec=True)
+    def test_scp_open_connection_socket_timeout_exception(self, mock_sshclient_connect):
+        """Check scp get_connection failing with socket timeout."""
+        from trollmoves.movers import ScpMover
+        import logging
+        import socket
+        import sys
```
mraspaud, line 162: These are in the standard library, can the be imported at the top of the file?

mraspaud (changes requested):
Thanks for adding this feature, makes a lot of sense to have the safeguard of a timeout!

trollmoves/movers.py
TAlonglong: Did not think of that. Good point.

trollmoves/tests/test_ssh_server.py
TAlonglong: yepp

pnuu (approved):
LGTM

PR #152, opened 2023-01-13
Do not transform local paths to ssh-uris

mraspaud:
For some reason in the past it was decided that files local to the server should be prefixed with "ssh://someserver" for publishing.
I can see a reason for doing this, and it triggers paramiko down the line in our setup (trollflow2).
So this PR removes this behaviour.

 - [x] Tests added

### DIFF
--- a/trollmoves/client.py
+++ b/trollmoves/client.py
@@ -491,7 +491,7 @@ def make_uris(msg, destination, login=None):
     """Create local URIs for the received files."""
     duri = urlparse(destination)
     scheme = duri.scheme
-    netloc = duri.netloc  # local file
+    netloc = duri.netloc
     if scheme != "s3" and duri.hostname and is_localhost(duri.hostname):
         scheme = ""
         netloc = ""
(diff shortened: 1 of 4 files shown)


trollmoves/client.py
```diff
@@ -485,20 +490,16 @@ def unpack_callback(var, **kwargs):
...
-    dest_hostname = duri.hostname or socket.gethostname()
-    if scheme not in ('s3') and socket.gethostbyname(dest_hostname) in get_local_ips():
-        scheme_, host_ = "ssh", dest_hostname  # local file
-    else:
-        scheme_, host_ = scheme, dest_hostname  # remote file
-        if login:
-            # Add (only) user to uri.
-            host_ = login.split(":")[0] + "@" + host_
+    scheme = duri.scheme
+    netloc = duri.netloc  # local file
```
pnuu, line 494: Not necessarily local.

pnuu (approved):
Just one comment, but LGTM

trollmoves/client.py
mraspaud: thanks

PR #153, opened 2023-01-20
Add retries to SCPMover.copy()

pnuu:
This PR adds retries to the `SCPMover.copy()` method. Currently there's only retries in creating the connection.

VERY early work.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -378,7 +378,7 @@ class ScpMover(Mover):
         _ = self._run_with_retries(self._copy, "SCP copy")
 
     def _copy(self):
-        from scp import SCPError
+        from scp import SCPException
 
         success = False
         try:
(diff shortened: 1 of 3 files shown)


trollmoves/movers.py
```diff
@@ -357,6 +357,43 @@ def move(self):
...
+        retries = 3
+        success = False
+        while retries > 0:
+            retries -= 1
+
+            try:
+                scp = self._get_scp_client()
+                scp.put(self.origin, self.destination.path)
+                success = True
+                break
```
TAlonglong, line 371: Do you need the break here if you break on success further down?

TAlonglong (commented):
Yeah I good with thisl.

Is it your intention to drop the timeout to the put function by doing this?

trollmoves/movers.py
pnuu: I was already pretty tired when writing that, so probably not :sweat_smile:

pnuu:
> Is it your intention to drop the timeout to the put function by doing this?

I think we can keep the timeout option, too.

PR #157, opened 2023-03-22
Add file creation time as metadata

mraspaud:
This PR adds the datetime metadata item `file_creation_time`, which is then available for the filepattern to use.

 - [x] Tests added  
 - [x] Fully documented

### DIFF
--- a/trollmoves/dispatcher.py
+++ b/trollmoves/dispatcher.py
@@ -165,7 +165,7 @@ from urllib.parse import urlsplit, urlunsplit, urlparse
 import yaml
 from posttroll.listener import ListenerContainer
 from posttroll.message import Message
-from posttroll.publisher import NoisyPublisher
+from posttroll.publisher import NoisyPublisher, create_publisher_from_dict_config
 from trollsift import compose
 
 from trollmoves.movers import move_it
@@ -190,7 +190,10 @@ def _create_publisher(publish_port, publish_nameservers):
 
 
 class Dispatcher:
-    """Class that dispatches files."""
+    """Class that dispatches files.
+
+    Idea for future refactoring: the publish arguments should really be provided in the configuration file.
+    """
 
     def __init__(self, config_file, publish_port=None, publish_nameservers=None, messages=None):
         """Initialize dispatcher class.
@@ -203,7 +206,11 @@ class Dispatcher:
 
         self.messages = messages
 
-        self.publisher = _create_publisher(publish_port, publish_nameservers)
+        if publish_port is not None:
+            self.publisher = PublisherReporter(self.config, publish_port, publish_nameservers)
+
+        else:
+            self.publisher = None
 
         self.host = socket.gethostname()
 
@@ -239,7 +246,7 @@ class Dispatcher:
             _check_file_locality(url, self.host)
             success = dispatch(url.path, destinations)
             if self.publisher:
-                self._publish(msg, destinations, success)
+                self.publisher.publish(msg, destinations, success)
 
     def get_destinations(self, msg):
         """Get the destinations for this message."""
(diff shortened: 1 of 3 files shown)


trollmoves/dispatcher.py
```diff
@@ -153,127 +157,53 @@
...
 from urllib.parse import urlsplit, urlunsplit, urlparse
-import socket
 
 import yaml
-import inotify.adapters
-from inotify.constants import IN_MODIFY, IN_CLOSE_WRITE, IN_CREATE, IN_MOVED_TO
 from posttroll.listener import ListenerContainer
-from posttroll.publisher import NoisyPublisher
 from posttroll.message import Message
+from posttroll.publisher import NoisyPublisher
```
pnuu, line 169: Not necessary here, but we really should be using `posttroll.publisher.create_publisher_from_dict_config` so that all publisher features were available.

mraspaud: This is now fixed along we more refactoring. Would be great if you could re-review :)

trollmoves/dispatcher.py
```diff
@@ -405,23 +285,114 @@ def create_dest_url(self, msg, client, conf):
...
-        if self.publisher:
+
+class PublisherReporter:
+    """This class uses a posttroll publisher to report the results of the dispatching.
+
+    This is the first of possibly many reporting classes, eg for monitoring or generating reports daily.
+
+    Idea for future refactoring: This could be made flexible enough to be used in other parts of trollmoves, eg in
+    move_it_client to report moved files. The main problem is the to pass the right configuration/topic for the messages
+    to be correct.
```
pnuu, line 296: This also should maybe be in an issue instead of in the docstring.

trollmoves/dispatcher.py
```diff
@@ -405,23 +285,114 @@ def create_dest_url(self, msg, client, conf):
...
+            "port": publish_port,
+            "nameservers": publish_nameservers
+        }
+
+        self._pub_starter = create_publisher_from_dict_config(pub_settings)
+        self.publisher = self._pub_starter.start()
+
+        # self.publisher = NoisyPublisher("dispatcher", port=publish_port,
+        #                                 nameservers=publish_nameservers)
+        # self.publisher.start()
```
pnuu, line 314: This should be deleted.
```suggestion
```

trollmoves/dispatcher.py
```diff
@@ -405,23 +285,114 @@ def create_dest_url(self, msg, client, conf):
...
+
+    topics = set()
+    for _client, client_config in new_config.items():
+        topics |= set(sum([item['topics'] for item in client_config['dispatch_configs']], []))
+
+    listener = ListenerContainer(topics=topics,
+                                 addresses=addresses,
+                                 nameserver=nameserver,
+                                 services=services)
+    yield listener
```
pnuu, line 394: Oooh, nice! This usage would have never crossed my mind.

trollmoves/dispatcher.py
```diff
@@ -405,23 +285,114 @@ def create_dest_url(self, msg, client, conf):
...
-        """Shutdown the dispatcher."""
-        logger.info('Terminating dispatcher.')
-        self.loop = False
-        try:
-            self.listener.stop()
-        except Exception:
-            logger.exception("Couldn't stop listener.")
-        if self.publisher:
+
+class PublisherReporter:
```
pnuu, line 289: Should the base class be eventually in Posttroll?

mraspaud: I don't think so, this one will be pretty specific to file moving I believe.

trollmoves/dispatcher.py
mraspaud: yes, context manager can be really nice sometimes :)

trollmoves/dispatcher.py
```diff
@@ -149,224 +153,100 @@
...
+        publisher = NoisyPublisher("dispatcher", port=publish_port,
+                                   nameservers=publish_nameservers)
+        publisher.start()
+        return publisher
+
+
+class Dispatcher:
+    """Class that dispatches files.
+
+    Idea for future refactoring: the publish arguments should really be provided in the configuration file.
```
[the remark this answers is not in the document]
mraspaud: I would say we can have it an issue also, and maybe give a link to the issue here?

trollmoves/dispatcher.py
mraspaud: Fixed

PR #163, opened 2023-04-11
Handle the distapcthing of a dataset (keeping the same filenames)

adybbroe:
This PR adds the possibility to dispatch datasets, A dataset is a set of files with uid's and full uri's. It splits the dataset in separate files and dispatch each file using the original file name and all files of the dataset is of course dispatched to the same destination.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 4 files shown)


adybbroe:
It says I should add my name to the `AUTHORS.md` but I don't see this file?

djhoese:
The template was probably copied from satpy.

pnuu (approved):
LGTM

PR #165, opened 2023-04-13
Make it possible to pass connection-parameters to the S3 mover

adybbroe:
This PR adds the possible to pass connection parameters when moving files to S3. By doing that it may be possible to specify the endpoint-url and the keys in the yaml file directly or indirectly by keeping those secret keys in a config file using profiles as described https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-a-configuration-file

I tried documenting this in the example dispatcher yaml file.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [x] Fully documented 

### DIFF
--- a/examples/dispatch.yaml
+++ b/examples/dispatch.yaml
@@ -50,3 +50,67 @@ target1:
         - area: omerc_bb
           daylight: '<30'
           coverage: '>50'
+
+target-s3-example1:
+  host: s3://my-fancy-bucket/
+  connection_parameters:
+    client_kwargs:
+      endpoint_url: 'https://minio-server.mydomain.se:9000'
+      verify: false
+    secret: "my-super-secret-key"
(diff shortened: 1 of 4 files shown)


pnuu (commented):
Any chance to:
- add a test?
- make this work with Trollmoves Server? (or does it already?)

adybbroe:
> Any chance to:
> 
>     * add a test?
> 
>     * make this work with Trollmoves Server? (or does it already?)

Ok, thanks, yes, will try look at adding a test! Not sure about the second one, but we will see...

mraspaud:
@pnuu  from what I see, the request manager passes its config directly to `move_it`, so it should already be supported in the move_it_server

pnuu (approved):
LGTM

PR #166, opened 2023-04-13
Error in destination path when dispatching to S3

adybbroe:
When using the dispatcher to upload files to an S3-bucket the destination path is wrong.

As an example: 

`mover = S3Mover('/path/to/local/file.txt', 's3://atms-dr/target-dir/file.txt')` will result in `s3://atms-dr/target-dir/file.txt/file.txt`

This PR seeks to solve this.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,18 @@ Additional required packages:
 Additional required packages:
 - ``s3fs``
 
+Special behaviour on destination filepath when using the S3Mover class:
+
+If the destination prefix (~filepath) has a trailing slash ('/') the original
+filename will be appended (analogous to moving a file from one directory to
+another keeping the same filename).
+
+If the destination prefix does not have a trailing slash the operation will be
+analougus to moving a file from one directory to a new destination changing the
+filename. The new destination filename will be the last part of the provided
+destination follwing the last slash ('/')
+
+
 ## s3downloader
 
 This module is able to download files from a s3 endpoint.
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ another keeping the same filename).
 If the destination prefix does not have a trailing slash the operation will be
 analougus to moving a file from one directory to a new destination changing the
 filename. The new destination filename will be the last part of the provided
-destination follwing the last slash ('/')
+destination following the last slash ('/').
 
 
 ## s3downloader
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ filename will be appended (analogous to moving a file from one directory to
 another keeping the same filename).
 
 If the destination prefix does not have a trailing slash the operation will be
-analougus to moving a file from one directory to a new destination changing the
+analogous to moving a file from one directory to a new destination changing the
 filename. The new destination filename will be the last part of the provided
 destination following the last slash ('/').
 
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -420,7 +420,7 @@ class S3Mover(Mover):
 
     The transfer is initiated by Trollmoves Client by having destination that starts with "s3://".
 
-    All the connection configurations and such are done using the `fsspec` configuration system:
+    All the connection configurations and such may be done using the `fsspec` configuration system:
 
     https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration
 
@@ -433,6 +433,27 @@ class S3Mover(Mover):
                 "key": "ACCESSKEY"
             }
         }
+
+    However, using the this procedure may not be useful if having several
+    endpoints/buckets with their own access/secret keys. Instead one can use
+    aws profiles (placed in `.aws/config`) to for instance set the
+    access/secret keys for various endpoints and then keep the actual url of
+    the endpoints in the yaml configuration (see examples/dispatch.yaml).
+
+    See documentation on profiles here:
+    https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-a-configuration-file
+
+
+    NB! Special behaviour on destination filepath:
+
+    If the destination prefix (~filepath) has a trailing slash ('/') the
+    original filename will be appended (analogous to moving a file from one
+    directory to another keeping the same filename).
+
+    If the destination prefix does not have a trailing slash the operation will
+    be analougus to moving a file from one directory to a new destination
+    changing the filename. The new destination filename will be the last part
+    of the provided destination follwing the last slash ('/').
 
     """
 
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -90,6 +90,7 @@ class Mover(object):
 
     def __init__(self, origin, destination, attrs=None):
         """Initialize the Mover."""
+        LOGGER.debug("destination = %s", str(destination))
         try:
             self.destination = urlparse(destination)
         except AttributeError:
@@ -441,12 +442,16 @@ class S3Mover(Mover):
             raise ImportError("S3Mover requires 's3fs' to be installed.")
         s3 = S3FileSystem()
         destination_file_path = self._get_destination()
+        LOGGER.debug('destination_file_path = %s', destination_file_path)
         _create_s3_destination_path(s3, destination_file_path)
+        LOGGER.debug('Before call to put: destination_file_path = %s', destination_file_path)
+        LOGGER.debug('self.origin = %s', self.origin)
         s3.put(self.origin, destination_file_path)
 
     def _get_destination(self):
         bucket_parts = []
         bucket_parts.append(self.destination.netloc)
+
         if self.destination.path != '/':
             bucket_parts.append(self.destination.path.strip('/'))
         bucket_parts.append(os.path.basename(self.origin))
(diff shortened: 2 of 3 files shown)


trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,30 @@ def test_s3_copy_file_to_base(S3FileSystem):
     S3FileSystem.return_value.put.assert_called_once_with(ORIGIN, "data-bucket/filename.ext")
 
 
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_get_destination_from_string(S3FileSystem):
+    """Test get the correct S3-bucket destination filename."""
+    s3_mover = _get_s3_mover(ORIGIN, "s3://data-bucket/")
+    result = s3_mover._get_destination()
+    assert result == 'data-bucket/filename.ext'
```
pnuu, line 97: Do not use the private method (and it should stay private, IMO). Instead check that the mocked `S3FileSystem.return_value.put()` has been called with the expected arguments.

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,30 @@ def test_s3_copy_file_to_base(S3FileSystem):
     S3FileSystem.return_value.put.assert_called_once_with(ORIGIN, "data-bucket/filename.ext")
 
 
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_get_destination_from_string(S3FileSystem):
```
pnuu, line 93: This should be 3 separate test cases.

The name is also too generic, also other tests passing strings to the mover. Maybe something like `test_s3mover_destination_FOO()` where `FOO` describes which input or feature is being tested.

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,30 @@ def test_s3_copy_file_to_base(S3FileSystem):
...
+
+    s3_mover = _get_s3_mover(ORIGIN, "s3://data-bucket/upload")
+    result = s3_mover._get_destination()
+    assert result == 'data-bucket/upload'
+
+
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_get_destination_from_parseresult(S3FileSystem):
+    """Test get the correct S3-bucket destination filename."""
+    s3_mover = _get_s3_mover(ORIGIN, urlparse("s3://data-bucket/upload/my_satellite_data.h5"))
```
pnuu, line 111: Hmm. The first thing that is done in the `Mover` base class is passing the `destination` to `urlparse()`. But I think this is a good to test if it isn't already.

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,30 @@ def test_s3_copy_file_to_base(S3FileSystem):
...
+    result = s3_mover._get_destination()
+    assert result == 'data-bucket/upload//my_satellite_data.h5'
```
pnuu, line 113: There should not be double `//` in the expected result. In S3 `foo/bar` and `foo//bar` are different objects, and we clearly want the single `/` in there. Also see the comment above about checking the `.put()` arguments instead of the private method.

pnuu (commented):
Some initial comments to the tests.

trollmoves/tests/test_movers.py
adybbroe: That was a typo, thx!

adybbroe:
@pnuu Thanks for the comments and suggestion. Think I have done what you requested, and made a suggested solution so tests are passing

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,33 @@ def test_s3_copy_file_to_base(S3FileSystem):
...
+    """Test get the correct S3-bucket destination filename."""
+    s3_mover = _get_s3_mover(ORIGIN, "s3://data-bucket/upload/")
+    s3_mover.copy()
+
+    S3FileSystem.return_value.put.assert_called_once_with(ORIGIN, "data-bucket/upload/filename.ext")
+
+
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_copy_file_to_prefix_no_trailing_slash(S3FileSystem):
+    """Test get the correct S3-bucket destination filename."""
```
pnuu, line 103:
```suggestion
    """Test giving destination without trailing slash to see it is used as object name."""
```

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,33 @@ def test_s3_copy_file_to_base(S3FileSystem):
...
+    s3_mover = _get_s3_mover(ORIGIN, "s3://data-bucket/upload")
+    s3_mover.copy()
+
+    S3FileSystem.return_value.put.assert_called_once_with(ORIGIN, "data-bucket/upload")
+
+
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_copy_file_to_prefix_urlparse_file_with_extention(S3FileSystem):
+    """Test get the correct S3-bucket destination filename."""
```
pnuu, line 112:
```suggestion
def test_s3_copy_file_to_prefix_urlparse(S3FileSystem):
    """Test that giving urlparse() result as destination works."""
```

trollmoves/tests/test_movers.py
```diff
@@ -88,6 +89,33 @@ def test_s3_copy_file_to_base(S3FileSystem):
     S3FileSystem.return_value.put.assert_called_once_with(ORIGIN, "data-bucket/filename.ext")
 
 
+@patch('trollmoves.movers.S3FileSystem')
+def test_s3_copy_file_to_prefix_with_trailing_slash(S3FileSystem):
+    """Test get the correct S3-bucket destination filename."""
```
pnuu, line 94:
```suggestion
    """Test that when destination ends in a slash, the original file basename is added to it."""
```

pnuu (commented):
Some suggested changes to test names and their docstrings.

I also think the difference between `s3://foo/bar` and `s3://foo/bar/` should be documented in the `S3Mover` class docstring and maybe also to `README.md` until we have proper documentation.

mraspaud:
> I also think the difference between `s3://foo/bar` and `s3://foo/bar/` should be documented in the `S3Mover` class docstring and maybe also to `README.md` until we have proper documentation.

I think I observed unexpected behaviour with the final / in the ftp mover, so maybe it's a more general problem?

pnuu (approved):
LGTM

pnuu:
Oh, the description of the difference between targets with and without a trailing slash is still needed.

pnuu (changes requested):
The difference between `s3://foo/bar` and `s3://foo/bar/` as a destination needs to be described.

adybbroe:
> Oh, the description of the difference between targets with and without a trailing slash is still needed.

Yes, forgot that, sorry, so thanks both for catching this. Should be done now. Also copied some text to the README.

README.md
```diff
@@ -138,6 +138,18 @@ Additional required packages:
...
+Special behaviour on destination filepath when using the S3Mover class:
+
+If the destination prefix (~filepath) has a trailing slash ('/') the original
+filename will be appended (analogous to moving a file from one directory to
+another keeping the same filename).
+
+If the destination prefix does not have a trailing slash the operation will be
+analougus to moving a file from one directory to a new destination changing the
+filename. The new destination filename will be the last part of the provided
+destination follwing the last slash ('/')
```
pnuu, line 150:
```suggestion
destination following the last slash ('/').
```

README.md
```diff
@@ -138,6 +138,18 @@ Additional required packages:
...
 - ``s3fs``
 
+Special behaviour on destination filepath when using the S3Mover class:
+
+If the destination prefix (~filepath) has a trailing slash ('/') the original
+filename will be appended (analogous to moving a file from one directory to
+another keeping the same filename).
+
+If the destination prefix does not have a trailing slash the operation will be
+analougus to moving a file from one directory to a new destination changing the
```
pnuu, line 148:
```suggestion
analogous to moving a file from one directory to a new destination changing the
```

trollmoves/movers.py
```diff
@@ -434,6 +434,27 @@ class S3Mover(Mover):
...
+
+
+    NB! Special behaviour on destination filepath:
+
+    If the destination prefix (~filepath) has a trailing slash ('/') the
+    original filename will be appended (analogous to moving a file from one
+    directory to another keeping the same filename).
+
+    If the destination prefix does not have a trailing slash the operation will
+    be analougus to moving a file from one directory to a new destination
```
pnuu, line 454:
```suggestion
    be analogous to moving a file from one directory to a new destination
```

trollmoves/movers.py
```diff
@@ -434,6 +434,27 @@ class S3Mover(Mover):
...
+    changing the filename. The new destination filename will be the last part
+    of the provided destination follwing the last slash ('/').
```
pnuu, line 456:
```suggestion
    of the provided destination following the last slash ('/').
```

pnuu (approved):
LGTM.

I fixed some typos.

PR #170, opened 2023-05-21
adding backup targets

### LINKED ISSUE #169: In the server add backup target host for scp
## Feature Request

Add backup or aditional target hosts using scp in server.

Some pseudo code suggested by @pnuu 
```
if transfer_failed and "backup_target" in config:
    for tgt in config["backup_target"]:
        status = retry_transfer(msg, target=tgt)
        if status:
            break
```

This will allow more than target hosts when using scp with the server which will increase redundancy I need for my setup.

Assign this to me and tag it for PCW

TAlonglong:
- [x] Closes #169  
 - [x] Tests added  
 - [x] Fully documented 

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2012-2020
+# Copyright (c) 2012-2023
 #
 # Author(s):
 #
@@ -296,25 +296,25 @@ class ScpMover(Mover):
     active_connections = dict()
     active_connection_lock = Lock()
 
-    def open_connection(self):
-        """Open a connection."""
+    def _ssh_connect(self, destination=None):
         from paramiko import SSHClient, SSHException
-
         retries = 3
         ssh_key_filename = self.attrs.get("ssh_key_filename", None)
         timeout = self.attrs.get("ssh_connection_timeout", None)
+        if not destination:
+            destination = self.destination.hostname
         while retries > 0:
             retries -= 1
             try:
                 ssh_connection = SSHClient()
                 ssh_connection.load_system_host_keys()
-                ssh_connection.connect(self.destination.hostname,
+                ssh_connection.connect(destination,
                                        username=self._dest_username,
                                        port=self.destination.port or 22,
                                        key_filename=ssh_key_filename,
                                        timeout=timeout)
                 LOGGER.debug("Successfully connected to %s:%s as %s",
-                             self.destination.hostname,
+                             destination,
                              self.destination.port or 22,
                              self._dest_username)
             except SSHException as sshe:
@@ -329,7 +329,22 @@ class ScpMover(Mover):
             ssh_connection.close()
             time.sleep(2)
             LOGGER.debug("Retrying ssh connect ...")
-        raise IOError("Failed to ssh connect after 3 attempts")
+
+    def open_connection(self):
+        """Open a connection."""
+        messg = ""
+        ssh_connection = self._ssh_connect()
+        if ssh_connection:
+            return ssh_connection
+        else:
+            #Failed to connect for some reason.
+            # Check if backup targets
+            for destination in self.attrs.get("backup_targets", []):
+                messg = " and backup targets"
+                ssh_connection = self._ssh_connect(destination)
+                if ssh_connection:
+                    return ssh_connection
+        raise IOError(f"Failed to ssh connect after 3 attempts{messg}.")
 
     @staticmethod
     def is_connected(connection):
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -309,7 +309,7 @@ class ScpMover(Mover):
             try:
                 ssh_connection = SSHClient()
                 ssh_connection.load_system_host_keys()
-                ssh_connection.connect(self.destination.hostname,
+                ssh_connection.connect(destination,
                                        username=self._dest_username,
                                        port=self.destination.port or 22,
                                        key_filename=ssh_key_filename,
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -296,19 +296,20 @@ class ScpMover(Mover):
     active_connections = dict()
     active_connection_lock = Lock()
 
-    def _ssh_connect(self, destination=None):
+    def open_connection(self):
+        """Open a connection."""
         from paramiko import SSHClient, SSHException
         retries = 3
         ssh_key_filename = self.attrs.get("ssh_key_filename", None)
         timeout = self.attrs.get("ssh_connection_timeout", None)
-        if not destination:
-            destination = self.destination.hostname
+        destination = self.destination.hostname
+        backup_targets = self.attrs.get("backup_targets", [])
         while retries > 0:
             retries -= 1
             try:
                 ssh_connection = SSHClient()
                 ssh_connection.load_system_host_keys()
-                ssh_connection.connect(destination,
+                ssh_connection.connect(self.destination.hostname,
                                        username=self._dest_username,
                                        port=self.destination.port or 22,
                                        key_filename=ssh_key_filename,
@@ -329,22 +330,11 @@ class ScpMover(Mover):
             ssh_connection.close()
             time.sleep(2)
             LOGGER.debug("Retrying ssh connect ...")
-
-    def open_connection(self):
-        """Open a connection."""
-        messg = ""
-        ssh_connection = self._ssh_connect()
-        if ssh_connection:
-            return ssh_connection
-        else:
-            #Failed to connect for some reason.
-            # Check if backup targets
-            for destination in self.attrs.get("backup_targets", []):
-                messg = " and backup targets"
-                ssh_connection = self._ssh_connect(destination)
-                if ssh_connection:
-                    return ssh_connection
-        raise IOError(f"Failed to ssh connect after 3 attempts{messg}.")
+            if retries == 0 and backup_targets:
+                destination = backup_targets.pop(0)
+                LOGGER.info("Changing destination to backup target: %s", destination)
+                retries = 3
+        raise IOError(f"Failed to ssh connect after 3 attempts.")
 
     @staticmethod
     def is_connected(connection):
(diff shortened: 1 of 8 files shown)


pnuu (approved):
LGTM. Just couple of small thoughts that can be ignored.

trollmoves/movers.py
```diff
@@ -329,7 +330,11 @@ def open_connection(self):
             ssh_connection.close()
             time.sleep(2)
             LOGGER.debug("Retrying ssh connect ...")
-        raise IOError("Failed to ssh connect after 3 attempts")
+            if retries == 0 and backup_targets:
+                destination = backup_targets.pop(0)
+                LOGGER.info("Changing destination to backup target: %s", destination)
+                retries = 3
+        raise IOError(f"Failed to ssh connect after 3 attempts.")
```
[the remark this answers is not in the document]
TAlonglong: Yeah, I was thinking something like that. You see I have the `f` I added and the forgot about it. I will a suggestion

trollmoves/tests/test_ssh_server.py
```diff
@@ -175,6 +175,32 @@ def test_scp_open_connection_socket_timeout_exception(self, mock_sshclient_conne
...
+        scp_mover = ScpMover(self.origin, self.destination_no_port,
+                             attrs={'ssh_connection_timeout': 1,
+                                    'backup_targets': ['backup_host1',
+                                                       'backup_host2']})
+        try:
+            with self.assertLogs(logger, level=logging.INFO) as lc, self.assertRaises(IOError):
+                scp_mover.open_connection()
+            self.assertIn(("SSH connection timed out:"), lc.output[0])
+            self.assertIn(("Changing destination to backup target: backup_host1"), lc.output[3])
+            self.assertIn(("Changing destination to backup target: backup_host2"), lc.output[7])
```
mraspaud, line 198: Now that we are using pytest, the preferred way to do this is
`assert "SSH connection timed out:" in lc.output[0]`

Also caplog could be used, but no rush for that.

mraspaud (approved):
LGTM, just one styling comment.

pnuu:
One thing I forgot to ask for: example configuration. Which made me realise that the original primary `destination` is configured in the Client side, but the backups need to be in the Server config, right? Should the backup servers be sent in the request message by the Client? That way the Server process could be used also for other transfers and it wouldn't fallback to push the data to wrong backup hosts.

TAlonglong:
Ah, crap. You are right.

TAlonglong:
The implementation I came up with does not work at all.

TAlonglong:
So, I think I will revert all of this work and start over. Maybe I'll start a new PR to add the `start()` to the publisher to be compatible with posttroll 1.10.0

Then I need a sanity check. I think I will implement the backup target if I hit here: https://github.com/pytroll/trollmoves/blob/main/trollmoves/client.py#L671

So the idea is then to, if backup targets/destinations config is provided, restart the request.

Any ideas?

pnuu:
I would keep the structure you have here, but instead of putting the backup destinations to Server config, put them in Client config and add them in the request message maybe in [this](https://github.com/pytroll/trollmoves/blob/main/trollmoves/client.py#L277) else block. Then in Server check the message for `backup_destinations` list in the message data if the retries go to zero.

TAlonglong:
Ah, yes. That looks doable. 

So, where is this tested then. Then I can try from there

TAlonglong:
Updated to have the backup targets config in the client config.
This is then sent to the server, which change the destination if the primary fails.
It's tricky to propagate this info from the message to the actual mover, so I added a extra function parameter for this.
One issue I see is that the final message contains the primary target, I have not found a way to replace that information in the message.

Please let me know id you see anything odd

pnuu:
> One issue I see is that the final message contains the primary target, I have not found a way to replace that information in the message.

This is a tricky one. The message should maybe be updated [here](https://github.com/pytroll/trollmoves/blob/main/trollmoves/server.py#LL129C15-L129C15), but that would require rather large changes so that the actual destination were passed down from the individual `Mover` to `movers.move_it()` to `server.RequestManager._move_file()` to `server.RequestManager._move_files()` and eventually to `movers.RequestManager.push()`. And then there's the complication that `_move_files()` already returns either an error message or `None`.

trollmoves/tests/test_client.py
```diff
@@ -167,6 +167,17 @@
...
 
+CLIENT_CONFIG_BACKUP_TARGETS = """
+[foo]
+providers = bar
+destination = scp://primary_host/tmp/foo
+login = user
+topic = /1b/hrit-segment/0deg
+publish_port = 0
+nameservers = ns1
+backup_targets=backup_host1 backup_host2
```
pnuu, line 178: These hosts need to have the same protocol as the original destination.

trollmoves/tests/test_client.py
```diff
@@ -917,6 +934,37 @@ def test_request_push_single_call(send_ack, send_request, clean_ongoing_transfer
...
+@patch('trollmoves.client.file_cache', new_callable=deque)
+@patch('trollmoves.client.clean_ongoing_transfer')
+@patch('trollmoves.client.send_request')
+@patch('trollmoves.client.send_ack')
+def test_request_push_backup_targets(send_ack, send_request, clean_ongoing_transfer, file_cache, ongoing_transfers):
+    """Test trollmoves.client.request_push() with a single file."""
+    from trollmoves.client import request_push
+    from tempfile import gettempdir
+
+    MSG_FILE_BACKUP_TARGETS = MSG_FILE2
```
pnuu, line 947:
```suggestion
    msg_file_backup_targets = MSG_FILE2
```
This should be lower case, and adjusted accordingly also later in the code.

pnuu (approved):
Just couple of small comments inline, otherwise LGTM.

The comment on the config made me think that *later* we could move the retry logic to `movers.move_it()` so that all of the movers would have the possibility. Then we could also have backup destinations with different transport schemes; the primary could be `scp://` and backups `scp://` and `ftp://` for example. But that's for another PR.

trollmoves/tests/test_client.py
TAlonglong: Yeah, and that reminds me of adding an example explaining this.

trollmoves/tests/test_client.py
TAlonglong: Hm, ok. I was just following the syntax used in the rest of the similar setup.

TAlonglong:
I'm quite happy with this. Please let me know what you think

trollmoves/movers.py
```diff
@@ -71,7 +71,12 @@ def move_it(pathname, destination, attrs=None, hook=None, rel_path=None):
         raise
 
     try:
-        mover(pathname, new_dest, attrs=attrs).copy()
+        m = mover(pathname, new_dest, attrs=attrs, backup_targets=backup_targets)
+        m.copy()
+        last_dest = m.destination
```
pnuu, line 76: Ah! Nice catch!

trollmoves/movers.py
```diff
@@ -329,7 +343,13 @@ def open_connection(self):
             ssh_connection.close()
             time.sleep(2)
             LOGGER.debug("Retrying ssh connect ...")
-        raise IOError("Failed to ssh connect after 3 attempts")
+            if retries == 0 and backup_targets:
+                backup_target = backup_targets.pop(0)
+                self.destination = self.destination._replace(netloc=f"{self.destination.username}@{backup_target}")
```
pnuu, line 348: I was going to say using private methods might not be a good idea, but apparently this is *public* API. From https://docs.python.org/3/library/collections.html#collections.namedtuple

> To prevent conflicts with field names, the method and attribute names start with an underscore.

trollmoves/movers.py
pnuu: ... although this only changes the log message, not the message published by the Client :see_no_evil:

pnuu: One possibility:
* in `movers.move_it()` return `m.destination`
* in `server.RequestManager._move_file()` replace `message.data['destination']` with the return value and return the modified message if the transfer was succesful
  * in the case of error do what is done already
* in `server.RequestManager._move_files()` after the `self._move_file()` call check `message.type is 'err'` instead of `error_message is not None`
* in `server.RequestManager.push()` again check the `message.type`, run `clean_url` if necessary

trollmoves/movers.py
TAlonglong: Yeah I also figured out that.

trollmoves/movers.py
TAlonglong: Ah, you are right. Crap.

TAlonglong:
A step further. What do you think?

trollmoves/server.py
```diff
@@ -129,28 +129,28 @@ def pong(self, message):
     def push(self, message):
         """Reply to push request."""
         new_msg = self._move_files(message)
-        if new_msg is None:
+        if new_msg.type is not 'err':
```
pnuu, line 132: The default return value is `None`, so this will fail with `AttributeError`. Maybe this is safer?
```suggestion
        if new_msg and new_msg.type is not 'err':
```

trollmoves/server.py
```diff
@@ -129,28 +129,28 @@ def pong(self, message):
...
+            _destination = clean_url(new_msg.data['destination'])
```
pnuu, line 133: Isn't the destination already cleaned in `_move_file()`? Or actually, maybe safer to do the cleaning here if the `err` message also has some credentials or such that needs to be cleaned.

pnuu (commented):
Couple more things. And a live test would be great :grin:

pnuu:
No need to demonstrate :joy: Just run a test that this works in real life.

pnuu (approved):
LGTM.

Does @mraspaud want to have a look?

TAlonglong:
Then I see something that might be a feature. First time a backup target is used, the active connection is tagged with the primary target. Therefore the next time, if the previous connection is still alive, this is used even if the actual connection is to the backup target. SO that way if this occur it looks like the primary target is used but the actual cache connection is to the bakcup target.

changing line https://github.com/pytroll/trollmoves/blob/main/trollmoves/movers.py#L136 to
`self.active_connections[(self.destination.hostname, port, username)] = connection, timer` will solve this, but I'm not sure how the behaviour should be

TAlonglong:
Not sure if this is ready to merge as I see the approval from @mraspaud is old/changes after the approval.

lahtinep:
I'm merging this, we need a new release with the `.start()` method to make PyPI installation work with the latest Posttroll.

PR #174, opened 2023-06-14
Fix ftp mover using destination filename when provided

mraspaud:
This PR fixes the ftp mover using the original filename even when a destination filename was provided.

 - [x] Tests added  
 - [ ] Fully documented


lahtinep (approved):
LGTM

pnuu (approved):
LGTM also from the correct account :roll_eyes:

mraspaud:
Thanks a lot for the fast review and merge!

PR #180, opened 2023-09-08
Delete error solved

bornagain1981:
On server.py the delete option accepts only boolean `False`, without quotes. An error occurred when you're working for the decompression of files on the fly.

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -946,6 +946,7 @@ def unpack(pathname,
             LOGGER.exception("Could not decompress %s", pathname)
         else:
             if delete in ["1", "yes", "true", "on"]:
+                delete = True
                 os.remove(pathname)
             return new_path
     return pathname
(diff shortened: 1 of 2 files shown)


trollmoves/server.py
```diff
@@ -945,7 +945,7 @@ def unpack(pathname,
         except Exception:
             LOGGER.exception("Could not decompress %s", pathname)
         else:
-            if delete.lower() in ["1", "yes", "true", "on"]:
+            if delete in ["1", "yes", "true", "on"]:
```
pnuu, line 948: This won't work if `delete` is set to any value that evaluates as `True`.

pnuu (changes requested):
This will need a unit test that checks the three options work:
* user doesn't define `delete` in the config -> the default is used
* user defines `delete = True` and the file is deleted
* user defines `delete = False` and the file is not deleted

The current implementation will fail on the `True` case.

bornagain1981:
@pnuu I traced all the options (1,yes..) back to True. If I use test_server.py on test folder I have no error message

PR #184, opened 2023-10-09
Make sure unpack takes `delete` as a boolean

mraspaud:
For some reason, the `delete` config item it converted to boolean, but the unpacking was never updated to take this into account. This PR fixes this.

 - [x] Tests added


TAlonglong (approved):
LGTM

pnuu (approved):

PR #185, opened 2023-10-12
Replace pyinotify with watchdog

### LINKED ISSUE #192: Migrate away from `pyinotify`
## Feature Request

The `pyinotify` library uses `asyncore` from Python stdlib, but it isn't available in 3.12. Also the `pyinotify` seems to be unmaintained, the last updates have been done 4 years ago.

**Describe the solution you'd like**
Migrate to `watchdog`.

**Describe any changes to existing user workflow**
Should be purely internal changes, apart from not needing to install `pyinotify` anymore.

**Additional context**
We already use `watchdog` in the library.

mraspaud:
This PR replaces pyinotify which has been unmaintained for a few years by watchdog

 - [x] Closes #192
 - [ ] Tests added  
 - [ ] Fully documented 


trollmoves/server.py
```diff
@@ -963,6 +941,6 @@ def parse_args(args=None):
                         help="Disable glob and handling of backlog of files at start/restart",
                         action='store_true')
     parser.add_argument("-w", "--watchdog", default=False, action="store_true",
-                        help="Use Watchdog instead of inotify")
+                        help="Use Watchdog polling instead of os-based notifying")
```
pnuu, line 944: Great addition, I was thinking about how to handle this switch.

pnuu (approved):
Great job! The code also seem way simpler without the original `inotify` version. I'll try to test this tomorrow, but if I can't find the time, just merge.

pnuu:
These are the failures I see locally some times, but not with every run:
```
FAILED trollmoves/tests/functional/test_move_it_server_client.py::test_simple_publishing_with_untarring - KeyError: 'dataset'
ERROR trollmoves/tests/functional/test_move_it_server_client.py::test_simple_transfer - AttributeError: 'NoisyPublisher' object has no attribute 'heartbeat'
FAILED trollmoves/tests/test_server.py::test_file_detected_with_inotify_is_published - assert 0 == 1
```

I think the last one is solved by adding a short sleep between starting the `MoveItServer` and writing the file. I'll run the test in a loop for a while and if I don't get any failures with the added sleep I'll push that.

pnuu:
Changing the Client config so that a non-noisy publisher is created (`publish_port = 40000` and `nameserver = False`) the heartbeat problem disappears completely.

pnuu:
The `NoisyPublisher` class in Posttroll does not have a `heartbeat()` method. It might be a fluke that sometimes that error is not seen. `NoisyPublisher._publisher.heartbeat()` exists, so I think we need to add a `.heartbeat()` method that calls `self._publisher.heartbeat(). Creating a PR and linking it here.

PR #190, opened 2024-02-16
Fix connection parameter handling

### LINKED ISSUE #189: S3 mover passes all config items to `S3FileSystem()`
**Describe the bug**
S3 mover passes all config items to `S3FileSystem()`. This causes problems further down in `aiobotocore.session.AioSession(**self.kwargs) with AioSession.__init__() got an unexpected keyword argument 'foo'` where `foo` is any unknown kwarg. My guess is that there was a `**kwargs` in there so the extra kwargs were just discarded without a trace, but now there are only a limited set of kwargs and the extra ones cause problems.

**To Reproduce**
Setup Trollmoves for S3 transfers.

**Expected behavior**
The S3 transfers should work.

**Actual results**
```python
[2024-02-15 12:09:49,959 ERROR    trollmoves.movers] Something went wrong during copy of /tmp/mdssftd_ter/S-LSA_-HDF5_LSASAF_MSG_MDSSFTD_MSG-Disk_202402151145 to s3://satellite-data-eumetcast-lsasaf-mdssftd/: AioSession.__init__() got an unexpected keyword argument 'delete'
[2024-02-15 12:09:49,961 DEBUG    trollmoves.movers]   File "/current/lib/python3.10/site-packages/trollmoves/movers.py", line 78, in move_it
    m.copy()
  File "/current/lib/python3.10/site-packages/trollmoves/movers.py", line 495, in copy
    _create_s3_destination_path(s3, destination_file_path)
  File "/current/lib/python3.10/site-packages/trollmoves/movers.py", line 520, in _create_s3_destination_path
    s3.mkdirs(destination_path)
  File "/current/lib/python3.10/site-packages/fsspec/spec.py", line 1498, in mkdirs
    return self.makedirs(path, exist_ok=exist_ok)
  File "/current/lib/python3.10/site-packages/fsspec/asyn.py", line 118, in wrapper
    return sync(self.loop, func, *args, **kwargs)
  File "/current/lib/python3.10/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File "/current/lib/python3.10/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
  File "/current/lib/python3.10/site-packages/s3fs/core.py", line 914, in _makedirs
    await self._mkdir(path, create_parents=True)
  File "/current/lib/python3.10/site-packages/s3fs/core.py", line 899, in _mkdir
    await self._call_s3("create_bucket", **params)
  File "/current/lib/python3.10/site-packages/s3fs/core.py", line 341, in _call_s3
    await self.set_session()
  File "/current/lib/python3.10/site-packages/s3fs/core.py", line 502, in set_session
    self.session = aiobotocore.session.AioSession(**self.kwargs)
aiobotocore.session.AioSession(**self.kwargs)` with AioSession.__init__() got an unexpected keyword argument 'foo'
```

**Screenshots**
-

**Environment Info:**
 - OS: [e.g. OSX, Windows, Linux]
 - Trollmoves Version: current `main`

**Additional context**
I'll handle this by adding a separate handling for S3 settings in a PR possibly tomorrow.

pnuu:
In https://github.com/pytroll/trollmoves/pull/165 `S3FileSystem()` were given all the "extra" kwargs given to the class. For Dispatcher this was only `connection_parameters`, but for Trollmoves Server this is the full configuration file. The kwarg handling in `S3FileSystem` and `aiobotocore.core.AioSession` are such that the values not consumed by the former are passed to the latter. This caused unrecognized kwargs from the Trollmoves Server config being passed to `AioSession` and as it doesn't have a catch-all `**kwargs` they cause it to crashed.

This PR changes the config handling in the following way:
* Dispatcher passes the full config to `trollmoves.movers.move_it()`
* `S3Mover.copy()` gets the dictionary `connection_parameters` and passes it's items to `S3FileSystem()`
* dictionaries can be given in Trollmoves Server .ini configuration

Note that numeric values given in the nested dictionary are not handled.

 - [x] Closes #189  
 - [x] Tests added  
 - [x] Fully documented

### DIFF
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -584,6 +584,7 @@ def _read_ini_config(filename):
         _parse_nameserver(res[section], cp_[section])
         _parse_addresses(res[section])
         _parse_delete(res[section], cp_[section])
+        res[section] = _create_config_sub_dicts(res[section])
         if not _check_origin_and_listen(res, section):
             continue
         if not _check_topic(res, section):
@@ -620,6 +621,34 @@ def _parse_delete(conf, raw_conf):
     val = raw_conf.getboolean("delete")
     if val is not None:
         conf["delete"] = val
+
+
+def _create_config_sub_dicts(original):
+    # Take a copy so we can modify the values if necessary
+    res = dict(original.items())
+    for key in original.keys():
+        parts = key.split("__")
+        if len(parts) > 1:
+            _create_dicts(res, parts, original[key])
+            del res[key]
+    return res
+
+
+def _create_dicts(res, parts, val):
+    cur = res
+    for part in parts[:-1]:
+        if part not in cur:
+            cur[part] = {}
+        cur = cur[part]
+    cur[parts[-1]] = _check_bool(val)
+
+
+def _check_bool(val):
+    if val.lower() in ["0", "false"]:
+        return False
+    elif val.lower() in ["1", "true"]:
+        return True
+    return val
 
 
 def _check_origin_and_listen(res, section):
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -32,6 +32,7 @@ import os
 import subprocess
 import tempfile
 import time
+import warnings
 from collections import deque
 from threading import Lock, Thread
 from configparser import ConfigParser
@@ -662,6 +663,9 @@ def _form_connection_parameters_dict(original):
         res["connection_parameters"] = {}
     for key in original.keys():
         if key in CONNECTION_CONFIG_ITEMS:
+            warnings.warn(
+                f"Consider using connection_parameters__{key} instead of {key}.",
+                category=UserWarning)
             res["connection_parameters"][key] = original[key]
             del res[key]
     return res
--- a/trollmoves/server.py
+++ b/trollmoves/server.py
@@ -64,6 +64,8 @@ file_cache = deque(maxlen=61000)
 file_cache_lock = Lock()
 START_TIME = datetime.datetime.utcnow()
 
+CONNECTION_CONFIG_ITEMS = ["connection_uptime", "ssh_key_filename", "ssh_connection_timeout", "ssh_private_key_file"]
+
 
 class RequestManager(Thread):
     """Manage requests."""
@@ -165,7 +167,8 @@ class RequestManager(Thread):
         return_message = None
         try:
             destination = move_it(pathname, message.data['destination'],
-                                  self._attrs, rel_path=rel_path,
+                                  self._attrs["connection_parameters"],
+                                  rel_path=rel_path,
                                   backup_targets=message.data.get('backup_targets', None))
             message.data['destination'] = destination
         except Exception as err:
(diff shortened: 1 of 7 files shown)


mraspaud (changes requested):
S3Mover is unfortunately not the only place where the connection parameters are used, eg the ScpMover also uses this. It is though not tested, so we can't see it from the CI, but these changes very likely break it.
Also, it feels backwards to pass the full config to the movers when only the connection parameters are needed. Instead, I think we need to modify the move_it_server to only pass the connection parameters to the movers.
Or is there a use case I missed when the full config is needed by the movers?

pnuu:
> S3Mover is unfortunately not the only place where the connection parameters are used, eg the ScpMover also uses this. It is though not tested, so we can't see it from the CI, but these changes very likely break it.

`ScpMover` does not use `connection_parameters` dictionary. It was actually only `Dispatcher` that collected this dictionary and passed it to the movers as extra kwargs whereas Trollmoves Server passes the whole config.

> Also, it feels backwards to pass the full config to the movers when only the connection parameters are needed. Instead, I think we need to modify the move_it_server to only pass the connection parameters to the movers. Or is there a use case I missed when the full config is needed by the movers?

In Trollmoves Server we've never had the connection parameters, which is a dictionary, because it uses ini configs. Thus all the items are in the main level of the configuration. Putting these various connection related items in a dictionary would need special item-by-item handling when combining them from the config.

pnuu:
Adjusted based on a short discussion on Slack huddle.

trollmoves/server.py
```diff
@@ -622,6 +627,46 @@ def _parse_delete(conf, raw_conf):
...
+
+
+def _form_connection_parameters_dict(original):
+    # Take a copy so we can modify the values if necessary
+    res = dict(original.items())
+    if "connection_parameters" not in res:
+        res["connection_parameters"] = {}
+    for key in original.keys():
+        if key in CONNECTION_CONFIG_ITEMS:
+            res["connection_parameters"][key] = original[key]
```
mraspaud, line 665: should we have a deprecation warning here? to motivate people to move to the __ syntax?

trollmoves/tests/test_movers.py
```diff
@@ -173,14 +173,14 @@ def test_s3_copy_file_to_base_using_connection_parameters(S3FileSystem):
     """Test copying to base of S3 bucket."""
     # Get the connection parameters:
     config = yaml.safe_load(test_yaml_s3_connection_params)
-    attrs = config['target-s3-example1']['connection_parameters']
+    attrs = config['target-s3-example1']
```
mraspaud, line 176: shouldn't we be passing only the connection parameters here?

mraspaud (commented):
Looks good! just a couple of comments.

trollmoves/server.py
pnuu: Hmm. Maybe :thinking: I'll see think about this and see how to document this in the first place.

pnuu: Warning added in https://github.com/pytroll/trollmoves/pull/190/commits/e158f3a1fc2fc7a15a6e3872cd0ca45589b90080

mraspaud (approved):
LGTM

PR #191, opened 2024-02-16
Update python CI versions to cover 3.10 - 3.12

pnuu:
Update CI Python versions to 3.10, 3.11 and 3.12.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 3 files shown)


pnuu:
Ok, next hurdle is `pyinotify` which uses `asyncore` that is not available in 3.12 stdlib. There is an [external package](https://github.com/simonrob/pyasyncore) supplying this, but I think migrating to `watchdog` is the way to go. I'll create another issue about this.

pnuu:
https://github.com/pytroll/trollmoves/issues/192

pnuu:
In short about test failures: https://github.com/pytroll/trollmoves/pull/185  needs to be merged before we can start testing on Python 3.12.

mraspaud (approved):
LGTM!

PR #193, opened 2024-02-28
Use log file in tmp_path instead of 'somefile'

pnuu:
Simple fix for the logging tests which were writing a `somefile` on the filesystem. Now we use `tmp_path` fixture instead and check that the new log file exists when it should.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Fully documented 


mraspaud (approved):
LGTM!

PR #198, opened 2024-04-22
Add the fetcher functionality and documentation on rtd

mraspaud:
This PR adds a pytroll-fetcher script and functionality to download files received through messages, and documentations pages for trollmoves to be hosted on readthedocs.

 - [x] Tests added  
 - [x] Fully documented

### DIFF
--- a/docs/source/fetcher.rst
+++ b/docs/source/fetcher.rst
@@ -36,6 +36,12 @@ An example config would be for this command line interface would be::
         nameserver: false
 
 
+The `publisher_config` and `subscriber_config` dictionaries in the config are passed straight to posttroll's
+:py:func:`~posttroll.publisher.create_publisher_from_dict_config` and
+:py:func:`~posttroll.subscriber.create_subscriber_from_dict_config` respectively, so refer to the corresponding
+documentation to see how to populate them.
+
+
 API
 ***
 
--- a/docs/source/fetcher.rst
+++ b/docs/source/fetcher.rst
@@ -8,7 +8,7 @@ Command line script
 *******************
 
 The pytroll-fetcher script listen to posttroll messages and reacts by downloading files to a given locations. The help
-command of the scripts gives::
+command of the script gives::
 
     usage: pytroll-fetcher [-h] [-c LOG_CONFIG] config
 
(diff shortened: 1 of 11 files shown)


docs/source/fetcher.rst
```diff
@@ -0,0 +1,43 @@
...
+---------------
+
+The pytroll-fetcher script and trollmoves.fetcher library are tools to fetch (download) files or objects from local or
+remote filesystems.
+
+Command line script
+*******************
+
+The pytroll-fetcher script listen to posttroll messages and reacts by downloading files to a given locations. The help
+command of the scripts gives::
```
adybbroe, line 11:
```suggestion
command of the script gives::
```

docs/source/fetcher.rst
```diff
@@ -0,0 +1,43 @@
...
+    Thanks for using pytroll-fetcher!
+
+An example config would be for this command line interface would be::
+
+    destination: /home/myuser/downloads/
+    publisher_config:
+        nameservers: false
+        port: 1979
+    subscriber_config:
+        addresses:
```
adybbroe, line 34: Could we perhaps add another example based on topics here if you have a watcher running on the server that already is sending posttroll messages?

adybbroe (approved):
Looks really nice to me! Thanks. Only one tiny correction.
But would be good to have some more eyes on it. @TAlonglong or @pnuu ?

TAlonglong (approved):
I had a quick look

LGTM, but I need to run it and test it a bit before I can see if it is doing what it is supposed to. Then eventually we can iterate on it.

docs/source/fetcher.rst
pnuu: Or point to Posttroll documentation to see which settings are available.

docs/source/fetcher.rst
```diff
@@ -0,0 +1,43 @@
...
+        - ipc://some_pipe
```
pnuu, line 35: Is this "inter-process communication" protocol? I didn't know Posttroll supported other than `tcp://`.

mraspaud: Yeah, tcp is hardcoded here and there in posttroll, but when passing addresses directly anything zeromq supports should work

docs/source/fetcher.rst
mraspaud: Fixed with links to the posttroll documentation

pnuu (approved):
LGTM

PR #199, opened 2024-05-27
Replace utcnow (and fix style along the way)

mraspaud:
This PR replaces usages of utcnow in trollmoves.
Ruff wasn't happy about some of the code, so style was modified accordingly.

### DIFF
--- a/bin/remove_it.py
+++ b/bin/remove_it.py
@@ -246,7 +246,7 @@ def clean_section(pub, section, conf, is_dry_run=True):
             kws[key] = int(info[key])
         except KeyError:
             pass
-    ref_time = dt.datetime.now(dt.UTC) - dt.timedelta(**kws)
+    ref_time = dt.datetime.now(dt.timezone.utc) - dt.timedelta(**kws)
 
     for template in templates:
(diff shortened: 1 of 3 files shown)


bin/remove_it.py
```diff
@@ -231,7 +246,7 @@ def clean_section(pub, section, conf, is_dry_run=True):
             kws[key] = int(info[key])
         except KeyError:
             pass
-    ref_time = datetime.utcnow() - timedelta(**kws)
+    ref_time = dt.datetime.now(dt.UTC) - dt.timedelta(**kws)
```
pnuu: `dt.UTC` is added in 3.11 so it won't work with Python 3.10 which we support. Until we drop 3.10 we should use `dt.timezone.utc` instead.

pnuu (changes requested):
For 3.10 support we need to use `dt.timezone.utc`.

bin/remove_it.py
mraspaud: Yep, fixed

pnuu (approved):
LGTM, thanks!

PR #216, opened 2025-04-23
Sanitize connection settings when using S3Mover

pnuu:
In https://github.com/pytroll/trollmoves/commit/b0570dcde2960a6d0ebb9a1a3d156e5de2795d97 I introduced a bug that triggers

```
AioSession.__init__() got an unexpected keyword argument 'ssh_key_filename'
```

when transfer is attempted to S3 object storage.

This PR adds connection parameter sanitation that only allows kwargs that are listed in `s3fs.S3FileSystem` docstring.

 - [x] Tests added

### DIFF
--- a/trollmoves/movers.py
+++ b/trollmoves/movers.py
@@ -514,11 +514,15 @@ class S3Mover(Mover):
 
     """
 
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM.

PR #217, opened 2025-04-27
Make it configurable what kind of file time modes should be used ehn removing files

adybbroe:
This PR adds the options to controll waht kind of file time modes (`st_ctime`, st_atime` ot `st_mtime`) should be used when cleaning files using the `remove_it.py` script.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/filescleaner.py
+++ b/trollmoves/filescleaner.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2025 Pytroll Developers
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utility functions for cleaning files and directories."""
+
+
+import datetime as dt
+import logging
+import os
+from glob import glob
+
+LOGGER = logging.getLogger("__name__")
+
+
+try:
+    from posttroll.message import Message
+except ImportError:
+    def Message(*args, **kwargs):
+        """Fake message object in case posttroll isn't installed."""
+        del args, kwargs
+
+
+class FilesCleaner():
+    """Files cleaner class to accomodate cleaning of files acording to configured rules."""
+
+    def __init__(self, publisher, section, info, dry_run=True):
+        """Initialize the class."""
+        self.pub = publisher
+        self.section = section
+        self.info = info
+        self.dry_run = dry_run
+        self.st_time = self.info.get('st_time', 'st_ctime')
+
+    def clean_dir(self, ref_time, pathname):
+        """Clean up a directory."""
+        section_files = 0
+        section_size = 0
+        LOGGER.info("Cleaning %s", pathname)
+        flist = glob(pathname)
+
+        for filename in flist:
+            if not os.path.exists(filename):
+                continue
+            try:
+                stat = os.lstat(filename)
+            except OSError:
+                LOGGER.warning("Couldn't lstat path=%s", str(filename))
+                continue
+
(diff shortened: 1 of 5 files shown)


trollmoves/filescleaner.py
```diff
@@ -0,0 +1,123 @@
...
+class FilesCleaner():
+    """Files cleaner class to accomodate cleaning of files acording to configured rules."""
+
+    def __init__(self, publisher, section, info, dry_run=True):
+        """Initialize the class."""
+        self.pub = publisher
+        self.section = section
+        self.info = info
+        self.dry_run = dry_run
+        self.st_time = self.info.get('st_time', 'st_ctime')
```
pnuu, line 47:
```suggestion
        self.stat_time_method = self.info.get('st_time', 'st_ctime')
```

I think the attribute name could be more descriptive. My suggestion might not be much better, though :sweat_smile: 

There is no `st_time` in the output of `os.lstat()`. Did you mean either `st_atime` or `st_mtime`?

trollmoves/filescleaner.py
```diff
@@ -0,0 +1,123 @@
...
+        for filename in flist:
+            if not os.path.exists(filename):
+                continue
+            try:
+                stat = os.lstat(filename)
+            except OSError:
+                LOGGER.warning("Couldn't lstat path=%s", str(filename))
+                continue
+
+            if dt.datetime.fromtimestamp(stat.__getattribute__(self.st_time), dt.timezone.utc) < ref_time:
```
pnuu, line 65:
```suggestion
            if dt.datetime.fromtimestamp(getattr(stat,self.st_time), dt.timezone.utc) < ref_time:
```
This is cleaner.

trollmoves/filescleaner.py
```diff
@@ -0,0 +1,123 @@
...
+
+LOGGER = logging.getLogger("__name__")
+
+
+try:
+    from posttroll.message import Message
+except ImportError:
+    def Message(*args, **kwargs):
+        """Fake message object in case posttroll isn't installed."""
+        del args, kwargs
```
pnuu, line 35: Posttroll is a hard requirement, no need for this try/except block.

trollmoves/tests/conftest.py
```diff
@@ -13,3 +32,49 @@ def pytest_collection_modifyitems(items):
             it for it in sorted_items if module_mapping[it] == module
         ]
     items[:] = sorted_items
+
+
+@pytest.fixture
+def fake_config_file(tmp_path):
```
pnuu, line 38: The config is not a fake, it actually exists, so maybe rename the fixture?

trollmoves/tests/conftest.py
```diff
@@ -13,3 +32,49 @@ def pytest_collection_modifyitems(items):
...
+    """Create a fake configuration file."""
+    file_path = tmp_path / TEST_BASIC_FILESCLEANER_CONFIG_FILENAME
+    with open(file_path, 'w') as fpt:
+        fpt.write(TEST_BASIC_FILESCLEANER_CONFIG)
+
+    yield file_path
+
+
+@pytest.fixture
+def fake_file_structure(tmp_path):
```
pnuu, line 48: Nothing really fake here either.

pnuu (changes requested):
One possible typo that affects the usage (`st_time` doesn't exist) and needs fixing. Also few other comments.

No strong opinions though, I haven't had use for this bit of Trollmoves.

trollmoves/filescleaner.py
adybbroe: Ok, thanks. Yes, maybe your suggestion is better. I will contemplate a bit more and decide... Perhaps others have thoughts?

trollmoves/filescleaner.py
```diff
@@ -0,0 +1,116 @@
...
+class FilesCleaner():
+    """Files cleaner class to accomodate cleaning of files acording to configured rules."""
+
+    def __init__(self, publisher, section, info, dry_run=True):
+        """Initialize the class."""
+        self.pub = publisher
+        self.section = section
+        self.info = info
+        self.dry_run = dry_run
+        self.stat_time_method = self.info.get('st_time', 'st_ctime')
```
pnuu, line 41: Should the setting name in `self.info` match the variable name? Might be more clear for the user. Later on changing the name would be a breaking change.

pnuu (approved):
LGTM.

Just one question inline.

trollmoves/filescleaner.py
adybbroe: Yes, good point. I think `stat_time_method` is better than `st_time`. So will change!

pnuu (approved):

PR #227, opened 2025-10-14
Convert to pyproject.toml and hatchling/hatch-vcs

mraspaud:
- [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 


pnuu (changes requested):
The scripts are missing from `pyproject.toml`.

Also the `netifaces-plus` could replace `netifaces` in this same PR.

pnuu (approved):
I *think* I browsed through all the files, which mostly had only the license header remove and quotes changed, and now all the functionality should be present, so LGTM.

PR #233, opened 2026-01-23
Use Posttroll socket functions

mraspaud:
This thus allows trollmoves to use posttroll security features: zap handshake, white-listing and curve authentification

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 ]
 dependencies = [
     "netifaces-plus",
-    "posttroll>=1.5.1",
+    "posttroll>=1.13.1",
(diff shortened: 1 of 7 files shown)


pnuu (approved):
As long as there are no backwards incompatibilities in using `tcp://` addresses, I'm fine with the changes even if I'm don't really understand what's happening :sweat_smile:

TAlonglong (approved):
The changes look OK to me. But why are you changing from the get_context to the backend zmq

TAlonglong:
Ah ok, so zmq has some build in security options

PR #234, opened 2026-02-02
Follow links in remove_it

mraspaud:
This PR allow following symlinks for cleaning

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/filescleaner.py
+++ b/trollmoves/filescleaner.py
@@ -55,7 +55,7 @@ class FilesCleaner():
         section_files = 0
         section_size = 0
         for pathname in glob(pathname_template):
-            for dirpath, _dirnames, _ in os.walk(Path(pathname).parent):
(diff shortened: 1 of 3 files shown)


trollmoves/filescleaner.py
gerritholl: I would document the return value here, in particular since you're changing it from boolean to filename or None.

gerritholl (approved):
Looks good to me, with the caveat that I'm not a trollmoves user so my review is not very deep.  Just one comment on documentation.

trollmoves/filescleaner.py
mraspaud: Thanks, good point!

pnuu (approved):
I don't personally use this feature, but can't spot anything wrong with it either.

TAlonglong (approved):
Similar here. I don't use this, but don't see anything wrong

Issue #92, opened 2021-06-18
Transfer files first to a temporary filename


pnuu:
The files should be transferred to a temporary filenames (to a temporary directory?) and then moved to the proper filename/location. This would help Posttroll agnostic scripts working with the files, so that they are not taken to use prematurely before the transfers have completed.

See also #91

pnuu:
This would need to be somehow configurable. If the Client doesn't have access to the the transferred file (e.g. an external ftp server or scp/ssh location) it can't rename the file.

pnuu:
Closed by #238

PR #225, opened 2025-09-23
Allow profile again in order to specify where to find the S3 credentials

adybbroe:
This PR seaks to re-introduce the `profile` keyword for the yaml configuration, in order to identify which aws profile to use when getting the S3 credentials.

 - [x] Closes #221 
 - [ ] Tests added  
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #238, opened 2026-04-25
Add option to transfer to temporary filename

### LINKED ISSUE #148: ScpMover to transfer to a temporary file and rename after completion
## Feature Request

There could be an option using scp to transfer files to transfer the file to a temporary filename which get renamed upon completion.

Possible scenarios is when for some reason the transfer is interrupted for any reason you might get a valid filename with incomplete content.
Another is that you could think you have a complete file and start reading, when it's not complete.

pnuu:
Add options for transferring files first to a temporary filename. 

This was implemented using Github Copilot (GPT-5 mini) with a lot of iteration and some manual refactoring.

 - [x] Closes #148 #92 
 - [x] Tests added  
 - [x] Fully documented


pnuu:
Ok, I think this is ready for initial review. I see there are still some refactoring that could be done.

trollmoves/movers.py
```diff
@@ -8,24 +8,34 @@
...
 
+from ._mover_utils import ensure_final_directory_for_rename, ensure_remote_dirs
+
 S3_ALLOWED_SETTINGS = ["anon", "endpoint_url", "key", "secret",
                        "token", "use_ssl", "s3_additional_kwargs", "client_kwargs",
                        "requester_pays", "default_block_size", "default_fill_cache",
                        "default_cache_type", "version_aware", "cache_regions",
                        "asynchronous", "config_kwargs", "kwargs", "session",
                        "max_concurrency", "fixed_upload_size",
-                       "profile"]
```
adybbroe, line 28: Why do we remove "profile" as allowed settings?

adybbroe: I re-introduced in #225 as we needed it

pnuu: Seems that it was removed in https://github.com/pnuu/trollmoves/commit/fe5828732c50dece835715dad0953e0a59f6abec by the feature added by AI. I don't know why it got removed, and went unnoticed as it was not caught by tests. I'll add a test that makes sure none of the current (+ profile) settings are not removed.

pnuu: Fixed in https://github.com/pytroll/trollmoves/pull/238/commits/2dc8694ad273253907042657b3ba2a0ad55ec2df

AGENTS.md
mraspaud, line 1: Could you add an little about TDD and clean code?

trollmoves/tests/test_movers_use_tmp.py
```diff
@@ -0,0 +1,498 @@
...
+    path = tmp_path / "source" / "data.txt"
+    path.parent.mkdir(parents=True)
+    path.write_text("hello atomic transfer")
+    return path
+
+
+# ===========================================================================
+# Group A – Mover.tmp_destination_for (pure static method, no I/O)
+# ===========================================================================
+
```
mraspaud: Not needed.

trollmoves/tests/test_movers_use_tmp.py
```diff
@@ -0,0 +1,498 @@
...
+    from trollmoves.movers import SftpMover
+    assert SftpMover.supports_atomic() is True
+
+
+def test_supports_atomic_base_mover_returns_false():
+    """The base Mover class returns False as a safe default."""
+    from trollmoves.movers import Mover
+    assert Mover.supports_atomic() is False
+    assert Mover.supports_atomic(attrs={"use_tmp_on_transfer": True}) is False
+
```
mraspaud: These tests check implementation, not functionality. Also if we decide to keep these, `parametrize` should be used.

trollmoves/movers.py
```diff
@@ -48,35 +58,72 @@ def move_it(pathname, destination, attrs=None, hook=None, rel_path=None, backup_
...
-                    pathname, str(fake_dest))
-    return m.destination
+        LOGGER.info("Successfully copied %s to %s", pathname, str(fake_dest))
+    return mover.destination
+
+
+def _create_mover(mover_cls, pathname, new_dest, attrs, backup_targets=None, tmp_dest=None):
+    if tmp_dest:
+        return mover_cls(pathname, tmp_dest, attrs=attrs, backup_targets=backup_targets)
+    return mover_cls(pathname, new_dest, attrs=attrs, backup_targets=backup_targets)
```
mraspaud: Why not implement the tmp name in the mover base class directly?

trollmoves/movers.py
```diff
@@ -108,30 +155,69 @@ def move(self):
...
 
+    @classmethod
+    def supports_atomic(cls, attrs=None):
+        """Return True if this mover class supports atomic tmp→final transfers.
+
+        The default is False (conservative). Subclasses that implement
+        finalize_atomic_transfer should override and return True (or, in
+        the case of S3Mover, inspect *attrs* to decide).
+        """
+        return False
```
mraspaud: Why not make this a property? attrs is coming for __init__ anyways right?

trollmoves/_mover_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    if not path:
+        return
+    dirname = path
+    if os.path.isfile(path) or os.path.splitext(path)[1]:
+        dirname = os.path.dirname(path) or "."
+    os.makedirs(dirname, exist_ok=True)
+
+
+def _ensure_remote_dirs_ftp(connection, parts):
```
mraspaud, line 27: The name of this function is hiding the fact that, on top of creating the dirs, the cwd is moved there.

trollmoves/_mover_utils.py
```diff
@@ -0,0 +1,141 @@
...
+            try:
+                connection.mkdir(current)
+            except OSError:
+                try:
+                    connection.mkdir(part)
+                except OSError:
+                    pass
+
+
+def ensure_remote_dirs(connection, path):
```
mraspaud, line 125: The name of this function is hiding the fact that, on top of creating the dirs, the cwd is moved there.

README.md
```diff
@@ -39,6 +39,14 @@ the messages it publishes, along with a json representation of a `fsspec`
...
 to use `fsspec` to read and process the remote files.
 
+#### Using initial temporary filenames
+
+To avoid exposing partially-uploaded files, movers can be configured to upload
+first to a temporary name and be renamed/activated only after the transfer
+completes. See [the mover section](#Using temporary-initial-filenames-in-transfers)
+and `examples/move_it_server.ini` for details and examples.
+
+
```
mraspaud, line 49: Can we also have a dedicated doc file for move_it where this comes in?

mraspaud (changes requested):
Thanks for taking up this challenging task!

A few comments inline and also:
- Docstrings must start with a verb, and should be strings, not comments (in the test files)
- Remove "Group *" comment headers, I don't thing they are needed.

trollmoves/movers.py
pnuu: If it were a property, the class would need to be instantiated first. In `move_it()` we call the `_get_tmp_destination()` before instantiation. I'll see if the suggestion to move the whole tmp stuff to the base class changes things. When I get that far...

trollmoves/movers.py
pnuu: This needs a deeper look that I have time for now, so I'll get back to it at a later time.

trollmoves/_mover_utils.py
pnuu: The side-effect is removed in 121f4c48c5a157eced7e1df46b50f3c1c410afab

trollmoves/_mover_utils.py
pnuu: The side-effect is removed in 121f4c48c5a157eced7e1df46b50f3c1c410afab

trollmoves/tests/test_movers_use_tmp.py
pnuu: How these are testing the implementation? As far as I understand the "atomicity" is intrinsic to the protocol.

Refactored to use `parametrize` in 121f4c48c5a157eced7e1df46b50f3c1c410afab (should have been a separate commit, but forgot).

pnuu:
I'll have a further look next week, if I have some spare time.

trollmoves/tests/test_movers_use_tmp.py
pnuu: Removed in c3ac89ea180d89a2ba8b48161c33cff1c2862ecc

trollmoves/movers.py
pnuu: Done in f06593a5c7e6f193251d12604a38e8489d92fe30

trollmoves/movers.py
pnuu: Converted to `@property` in 30b81849f2f4344b4013685d49df72b5df1207d3

README.md
pnuu: Documentation added in 96f9f2d7a62aed276e8fb8bfbb0e93485b268e23

AGENTS.md
pnuu: Added in db6cf7d87174a7ab5e6cbdf3be5f9e2b46335fe8

The prompt was `add TDD and clean-code practices to AGENTS.md. The development should follow red-green-refactor cycle`.

pnuu:
I asked Claude to review the changes. Fixes coming.

pnuu:
Summary of the findings, fixes and their commit short hashes:

* 6ed1539: S3 atomic transfers died after the first file. Mover.__init__ held the caller's attrs by reference and _sanitize_attrs del'd from it — Trollmoves Server reuses one connection_parameters dict, so use_tmp_on_transfer was stripped from the live config on first transfer. Copy attrs in the base mover; add the key to the allow-list.
* cc60fd4: Multipart uploads renamed real files. _multipart_upload guessed the final key by stripping tmp_prefix from the basename, so .hidden.nc uploaded as hidden.nc. Pass the final destination in instead; _get_destination now takes an optional destination.
* ff34776: Directory destinations lost the filename. s3://bucket/dir/ + atomic transfer produced the tmp key bucket/dir/. and finalized to bucket/dir/. Resolve the origin filename before deriving the tmp destination.
* 4d20d98: Backup-target host wasn't reported. Restored the destination update in move_it (hook + log), and finalize now uses the netloc the transfer actually used — previously the published destination pointed clients at the primary host after a failover.
* 5c1bac0: Remote temp files leaked; local files at shadowing paths were deleted. Cleanup was a bare local os.remove for every scheme. Now dispatched through a remove_tmp_file hook implemented per protocol; cleanup errors are logged, never masking the transfer error.
* a90b610: AGENTS.md described a repo that doesn't exist — setup.py, setup.cfg, flake8, versioneer, bin/. Rewritten for pyproject/hatchling/ruff.
* 9fefc20: boto3 undeclared despite being required by the documented-preferred multipart path. Added to the s3 and all groups.
* 6815c1f: Removed the unused ensure_local_dir.
* 837d4da: Directory-creation failures were swallowed by four helpers, turning a permission-denied MKD into a confusing STOR error. They raise now; ensure_final_directory_for_rename was a near-duplicate and delegates.
* e1b2103: Fallback log downgraded to warning (matching the docs); restored the discarded-S3-key debug line (it was broken — a bare {str(key)}).
* 9bc5e50: Fixed the malformed README anchor; docs now state where temp files are cleaned up.

pnuu:
Found just one thing to fix. The sandbox conteiner I'm running Claude with didn't have SSH Server in it so Claude decided to add skipping those tests to `AGENTS.md`.

PR #241, opened 2026-06-18
Make sure never to delete configured base dir even if old

adybbroe:
Using the `remove_it.py` script we have experienced that using the recursive flag to clean directory structures even the old and empty base directory gets deleted. This is an unwanted feature! This PR fixes this. So I added a new test, where I could reproduce this error with the old code. See the new unit test: `test_remove_old_files_but_not_basedir`.

NB! Code updated is in the `filescleaner.py` and thus should only impact the `remove_it.py` part of Trollmoves.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/filescleaner.py
+++ b/trollmoves/filescleaner.py
@@ -56,17 +56,20 @@ class FilesCleaner():
         section_size = 0
         removed = []
 
-        for pathname in glob(pathname_template):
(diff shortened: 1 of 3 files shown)


pnuu (approved):
LGTM

mraspaud (approved):
LGTM!

PR #243, opened 2026-06-18
Handle cases with wildcards inside the path to the basedir-template

### LINKED ISSUE #242: Error handling templates like "somedir/otherdirs-??/*.ext" in remove-script
**Describe the bug**
The recent change to `filescleaner.py` in #241 fails to properly handle the cases when the configures base-dir template looks like this: "somedir/otherdirs-??/*.ext".

We need to inlcude a test case that catch this and then solve it.

**To Reproduce**

```python
# Your code here

```

**Expected behavior**

**Actual results**
Text output of actual results or error messages including full tracebacks if applicable.

**Screenshots**

**Environment Info:**
 - OS: [e.g. OSX, Windows, Linux]
 - Trollmoves Version: [e.g. 0.9.0]

**Additional context**

adybbroe:
We realised there was a bug in the recently merged PR = https://github.com/pytroll/trollmoves/pull/241

The changes implemnted to `filescleaner.py` in https://github.com/pytroll/trollmoves/pull/241 fails to properly handle the cases when the configures base-dir template looks like this: "somedir/otherdirs-??/*.ext".

That was apprently not acptured by existing unit tests. This is now repaired and the code updated/fixed.

 - [ ] Closes #242 
 - [x] Tests added  
 - [ ] Fully documented 

### DIFF
--- a/trollmoves/filescleaner.py
+++ b/trollmoves/filescleaner.py
@@ -56,20 +56,24 @@ class FilesCleaner():
         section_size = 0
         removed = []
 
-        base_path = Path(pathname_template).parent
-        pattern = Path(pathname_template).name
+        base_template = str(Path(pathname_template).parent)
+        file_pattern = Path(pathname_template).name
 
-        for dirpath, _dirnames, _ in os.walk(base_path, followlinks=True):
-            files_in_dir = glob(os.path.join(dirpath, pattern))
+        for base_dir in glob(base_template):
+            if not os.path.isdir(base_dir):
+                continue
 
-            s_size, s_files, removed_files = self.clean_files_and_dirs(files_in_dir, ref_time)
-            section_files += s_files
-            section_size += s_size
-            removed.extend(removed_files)
+            for dirpath, _dirnames, _ in os.walk(base_dir, followlinks=True):
+                files_in_dir = glob(os.path.join(dirpath, file_pattern))
 
-            # Do not remove the configured/base directory itself
-            if Path(dirpath) != base_path and not os.listdir(dirpath):
-                self._remove_empty_directory(dirpath)
+                s_size, s_files, removed_files = self.clean_files_and_dirs(files_in_dir, ref_time)
+                section_files += s_files
+                section_size += s_size
+                removed.extend(removed_files)
+
+                # Do not remove the configured/base directory itself
+                if Path(dirpath) != Path(base_dir) and not os.listdir(dirpath):
+                    self._remove_empty_directory(dirpath)
 
         return (section_size, section_files, removed)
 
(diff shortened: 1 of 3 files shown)


trollmoves/filescleaner.py
```diff
@@ -56,20 +56,24 @@ def clean_dir(self, ref_time, pathname_template, **kwargs):
         section_size = 0
         removed = []
 
-        base_path = Path(pathname_template).parent
-        pattern = Path(pathname_template).name
+        base_template = str(Path(pathname_template).parent)
+        file_pattern = Path(pathname_template).name
```
pnuu, line 61: Why only the first needs `str()`?

trollmoves/filescleaner.py
```diff
@@ -56,20 +56,24 @@ def clean_dir(self, ref_time, pathname_template, **kwargs):
...
+                section_files += s_files
+                section_size += s_size
+                removed.extend(removed_files)
 
-            # Do not remove the configured/base directory itself
-            if Path(dirpath) != base_path and not os.listdir(dirpath):
-                self._remove_empty_directory(dirpath)
+                # Do not remove the configured/base directory itself
+                if Path(dirpath) != Path(base_dir) and not os.listdir(dirpath):
+                    self._remove_empty_directory(dirpath)
```
pnuu: To reduce nesting, and to clarify what is happening, could this inner loop be a separate method?

trollmoves/tests/conftest.py
```diff
@@ -86,22 +86,27 @@ def file_structure_with_some_old_files_and_empty_dir(tmp_path):
...
     oldfile = (data_subdir1 / "b.png")
     os.utime(oldfile, (eight_hours_ago.timestamp(), eight_hours_ago.timestamp()))
+    oldfile = (data_subdir2 / "c.tif")
+    os.utime(oldfile, (eight_hours_ago.timestamp(), eight_hours_ago.timestamp()))
     # Force the directory to be old as well:
-    os.utime(data_subdir2, (eight_hours_ago.timestamp(), eight_hours_ago.timestamp()))
+    os.utime(data_subdir3, (eight_hours_ago.timestamp(), eight_hours_ago.timestamp()))
 
-    return data_dir, data_subdir1, data_subdir2
+    return data_dir, data_subdir1, data_subdir2, data_subdir3
```
pnuu: Could these have more descriptive names? Both here and where they are used?

pnuu (commented):
Some comments inline. I don't have strong objections, so feel free to merge if you feel this is good enough.

trollmoves/filescleaner.py
adybbroe: The latter always return a string, while `Path(pathname_template).parent)` is a `pathlib.PosixPath` class object

trollmoves/tests/conftest.py
adybbroe: Yes, you are right. Fixed in the latest push

trollmoves/filescleaner.py
adybbroe: I have refactored a bit now

